PR #65, opened 2026-04-30
Add manual email/password registration alongside OAuth - Upgrade PostgreSQL 16 → 18

### LINKED ISSUE #62: Add manual email/password registration alongside OAuth
## Why

OAuth-only sign-in is a hard gate for local testing and for users who don't want to bind a Discord/Google account just to try the platform. The Phase 1 deliverable is "log in → upload → see feed → like" — adding an email/password path keeps that loop usable in a no-third-party-creds setup (CI, contributor laptops, demo accounts).

This complements (not replaces) OAuth: the existing Discord/Google flows stay as-is.

## Server scope (`area:server`)

- New tables: `user_credentials` (or extend `users` with optional `password_hash` + `password_algo` columns).
  - Use Argon2id (preferred) or bcrypt — pick one, document the choice. Reject plaintext or fast hashes (SHA-256, MD5).
  - `password_hash` is optional on `users` so OAuth-only accounts don't need a row in `user_credentials`.
- Endpoints (mirroring existing auth surface in [server/src/GankedTV.Api/Endpoints/AuthEndpoints.cs](server/src/GankedTV.Api/Endpoints/AuthEndpoints.cs)):
  - `POST /auth/register` — body: \`{ email, username, password }\`. Validates uniqueness, hashes password, issues JWT + refresh token like the OAuth callback does.
  - \`POST /auth/login\` — body: \`{ email, password }\`. Verifies hash, issues JWT + refresh token.
  - Optional: \`POST /auth/password/change\` for authenticated users.
- Validation rules:
  - Email format + uniqueness (already enforced for OAuth users).
  - Username slug rules already exist — reuse \`UsernameGenerator.Slugify\`.
  - Password policy: minimum length (12+?), reject the most common breach-list passwords if cheap. Document the chosen rules.
- Rate limiting on \`/auth/login\` to slow credential-stuffing — at minimum per-IP, ideally per-email-target too.
- Migration: nullable \`password_hash\` + \`password_algo\` on \`users\` (or new table — pick one based on cleanest schema).

## Web scope (\`area:web\`)

- Add a \`/register\` and \`/login\` route alongside the existing OAuth buttons (current sign-in lives in [web/src/views/SignInView.vue](web/src/views/SignInView.vue) — verify path).
- Forms: email + username + password (register), email + password (login). Surface server validation errors inline.
- Auth store integration ([web/src/stores/auth.ts](web/src/stores/auth.ts)) — same JWT/refresh handling as OAuth, just a different acquisition path.

## Seed scope

- Extend \`SeedCommand\` ([server/src/GankedTV.Api/Tools/SeedCommand.cs](server/src/GankedTV.Api/Tools/SeedCommand.cs)) so the seeded \`seeduser\` (or a second \`testuser\`) has a known password hash, e.g. \`testpass123!\`. This lets contributors hit \`/auth/login\` directly after \`make seed\` instead of needing to call \`/dev/token\`.
- Document the credentials in the README "Local services" section.

## Out of scope

- Email verification flow (send email + click-to-confirm). Defer to a follow-up — not on the Phase 1 critical path.
- Password reset via emailed link. Same — follow-up.
- Account linking (\`oauth + password on the same row\`). Follow-up.

## Acceptance

- [ ] Migration adds password storage; coverage gate (85/85) holds.
- [ ] \`POST /auth/register\` and \`POST /auth/login\` work end-to-end against MinIO + Postgres.
- [ ] \`/register\` and \`/login\` pages let a fresh visitor create an account and sign in without leaving localhost.
- [ ] \`make seed\` lands a test account whose password is documented in the README, and \`POST /auth/login\` against that account returns a valid JWT.
- [ ] Existing OAuth flows still pass their integration tests.

### LINKED ISSUE #64: Upgrade PostgreSQL 16 → 18 (dev compose + test fixture)
## Why

We're on `postgres:16` for local dev and `postgres:16-alpine` for the test fixture. Postgres 18 is GA and brings real perf wins (async I/O on the read path, planner improvements, better partitioning) plus the usual security/support runway. While the data we have locally is throwaway seed data, this is the cheapest moment to bump.

## What

Bump both surfaces:

- **[`docker-compose.dev.yml`](docker-compose.dev.yml)** — `postgres:16` → `postgres:18`.
- **[`server/tests/GankedTV.Api.Tests/TestSupport/PostgresFixture.cs`](server/tests/GankedTV.Api.Tests/TestSupport/PostgresFixture.cs)** — `postgres:16-alpine` → `postgres:18-alpine`.

### Footgun: PGDATA path changed in postgres:18

The official `postgres` image moved `PGDATA` from `/var/lib/postgresql/data` (≤17) to `/var/lib/postgresql/18/docker` in the 18 image to make multi-version installs cleaner. Our compose currently mounts the named volume at the **old** path:

```yaml
volumes:
  - postgres_data:/var/lib/postgresql/data
```

Without a fix, PG18 will init a fresh DB at the new path inside the container layer and the `postgres_data` volume will silently end up empty.

Two acceptable fixes:

1. **Update the mount target** (preferred — moves with the upstream default):
   ```yaml
   volumes:
     - postgres_data:/var/lib/postgresql/18/docker
   ```
2. **Pin `PGDATA` back to the old path** (no mount change):
   ```yaml
   environment:
     PGDATA: /var/lib/postgresql/data
   ```

Either is fine for this project — local data is seed only, so a clean wipe via `make clean && make up` is the actual "migration."

The Testcontainers test fixture is unaffected: it uses ephemeral container storage with no host-mounted volume.

## Verification

- `make clean && make up && dotnet test server` — full suite green against PG18 in both surfaces (host compose for any manual checks, Testcontainers for the fixture).
- `pg_isready` health check still passes.
- `pgcrypto` extension is still available on PG18 (verified in [PostgresFixture.cs:31](server/tests/GankedTV.Api.Tests/TestSupport/PostgresFixture.cs#L31) — extension is created idempotently). No code change expected here, just confirming the fixture init still works.
- EF migrations apply cleanly — they go through Npgsql which is version-agnostic.

## Notes

- Caught while reviewing the storage-test PR (#63) — cheap to ride the upgrade train while we're already touching the test infrastructure.
- The data we have now is just dev seed data, so we don't need a `pg_upgrade` plan or volume migration script.

Turbootzz:
## Summary

Bundles two Phase 1 — Core Loop tickets into one PR because they share infra surface and want the same integration-test pass: bumps Postgres 16 → 18 and ships manual email/password auth alongside the existing OAuth providers, so the log in → upload → see feed → like loop is usable in CI / contributor-laptop / demo-account setups without third-party creds. Also adds a `make ci` developer-convenience target that mirrors the GitHub workflows.

Closes #64
Closes #62

## What's here

- **Postgres 18** in dev compose + Testcontainers fixture. `PGDATA` pinned to `/var/lib/postgresql/data/pgdata` (subdir of the existing volume) so the data path is stable across future major-version bumps — the upstream default moves per-major.
- **Email/password endpoints** mirroring the OAuth surface: `POST /auth/register`, `POST /auth/login`, `POST /auth/password` (authenticated, handles both first-time set and rotation). Argon2id (Konscious, OWASP baseline parameters, PHC encoding so cost factors can rotate later). Per-IP `FixedWindowLimiter` of 5/min on register + login. Generic 401 on login with constant-time-equivalent dummy verify on missing-user / no-password paths.
- **Account-merge policy.** OAuth callback's existing `TryLinkByEmailAsync` already covers OAuth↔OAuth and OAuth→password merging by verified email — kept as-is, regression test added. Password→OAuth direction is unsafe without email verification (deferred per the issue), so `/auth/register` returns 409 + nudges users to OAuth-then-attach via `POST /auth/password`.
- **Schema:** nullable `password_hash` + `password_algo` columns on `users` (extending vs separate table, per chat). Migration `AddUserPasswordCredentials`.
- **`/auth/me` now exposes `hasPassword: bool`** so the web settings view can switch between "Set password" and "Change password" copy correctly.
- **Web**: email/password form on `/login` alongside OAuth buttons, new `/register` view, new `/settings/password` view for OAuth users to attach a password.
- **Seed**: `seeduser@dev.local` / `testpass123!` documented in README. Idempotent — won't stomp on a rotated password.
- **`make ci` / `ci-server` / `ci-web`** mirror the GitHub workflow steps. Pre-push hook left untouched per direction (it scopes to changed dirs).
- **Coverage provider switched v8 → istanbul** in `vitest.config.ts` so `make ci-web` works under bun 1.3.13+ (bun's `node:inspector` polyfill doesn't implement the precise-coverage API v8 needs). CI numbers unchanged in shape.

## How to test manually

**One-time after pulling** (the PG18 PGDATA path move means existing local volumes need a wipe):

```bash
make clean && make up && make migrate && make seed
docker exec -it gankedtv-postgres-1 psql -U gankedtv -d gankedtv -c 'SELECT version();'  # expect PostgreSQL 18.x
```

**Full CI sweep**: `make ci` (server format/build/test+coverage, then web).

**Auth roundtrip:**

```bash
# Register
curl -X POST localhost:5000/auth/register -H 'content-type: application/json' \
  -d '{"email":"a@b.dev","username":"alice","password":"correct-horse-battery"}'

# Sign in as the seeded account
curl -X POST localhost:5000/auth/login -H 'content-type: application/json' \
  -d '{"email":"seeduser@dev.local","password":"testpass123!"}'

# Hit a protected route with the returned token
curl localhost:5000/auth/me -H "authorization: bearer <token>"

# Rate limit kicks in on the 6th login in 60s
for i in (seq 6); curl -s -o /dev/null -w "%{http_code}\n" -X POST localhost:5000/auth/login \
  -H 'content-type: application/json' -d '{"email":"x@y.z","password":"wrong-12345678"}'; end
# → 401, 401, 401, 401, 401, 429
```

**Account-merge sanity:**
- Register `b@c.dev` via `/auth/register`, then sign in via Discord/Google with a verified email matching `b@c.dev` → same user row, both `password_hash` and the provider id populated.
- Sign in via OAuth, then `POST /auth/password` with the access token to attach a password → `/auth/login` with that email succeeds and returns the same user.
- `POST /auth/register` with an email that already belongs to an OAuth-only account → 409 with `email_taken`.

**Web flow** (`make up && make server` + `make web`):
- `/register` → sign up → land on home authenticated.
- `/login` → sign in with `seeduser@dev.local` / `testpass123!` (or your registered account).
- OAuth buttons on `/login` still work — no regression.
- `/settings/password` switches to "Set password" copy when the account has no password (OAuth-only).

## Checklist

- [ ] One-time `make clean && make up && make migrate && make seed` run after pulling (PGDATA path moved).
- [ ] `make ci` passes locally (server 416/416 at 94/85, web 95/95 at 93/96).
- [ ] OAuth roundtrip still works after merging (regression-tested by the existing `AuthEndpointsCallbackTests` and a new `UserUpsertService` link-preserves-password test).

## Summary by CodeRabbit

* **New Features**
  * Email/password registration, credential login, and password set/change flows added.
  * UI: registration page, email/password sign-in option, and password settings page.
  * Login endpoint rate limiting to reduce abuse.
  * Stronger password rules enforced (longer minimum, common-password rejection).

* **Documentation**
  * Added local Makefile targets and README guidance to run full CI locally; seeded dev user credentials documented.

* **Tests**
  * Extensive auth, password policy, and rate-limit integration/unit tests added.

* **Chores**
  * Development Postgres image bumped to v18.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -42,6 +42,14 @@ make hooks                    # One-time: install pre-push hook via core.hooksPa
 ```
 Hook lives at `.githooks/pre-push` and mirrors CI, scoped to changed top-level dirs (`server/` or `web/`). Bypass with `PREPUSH_SKIP=1 git push` or `git push --no-verify`.
 
+### CI mirror (run all CI checks locally)
+```bash
+make ci                       # Runs server + web CI checks (format, build, test+coverage)
+make ci-server                # Server-only subset
+make ci-web                   # Web-only subset
+```
+Mirrors `.github/workflows/server.yml` and `.github/workflows/web.yml` step-for-step. Pre-push hook stays separate — it scopes to changed dirs; `make ci` is the full matrix.
+
 ### Server (from repository root)
 ```bash
 dotnet build server           # Build
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: up down clean logs server server-build server-test migrate migrate-add seed web web-install web-build web-test web-lint dev-all hooks
+.PHONY: up down clean logs server server-build server-test migrate migrate-add seed web web-install web-build web-test web-lint dev-all hooks ci ci-server ci-web
 
 # Infrastructure
 up:
@@ -64,6 +64,26 @@ dev-all: up
 	@trap 'kill 0' EXIT; \
 	dotnet watch --project server/src/GankedTV.Api & \
 	cd web && bun dev
+
+# CI mirror: runs the same checks the GitHub workflows run, in verify-only mode.
+# Mirrors `.github/workflows/server.yml` and `.github/workflows/web.yml` step-for-step
+# so contributors can reproduce CI locally with one command.
+#
+# Pre-push hook is intentionally NOT wired through here — it scopes to changed dirs
+# (server/ vs web/), `make ci` runs the full matrix. Use sub-targets for partial runs.
+ci: ci-server ci-web
+
+ci-server:
+	cd server && dotnet format --verify-no-changes
+	cd server && dotnet build --configuration Release
+	cd server && dotnet test --configuration Release /p:CollectCoverage=true /p:Threshold=85%2C85
+
+ci-web:
+	cd web && bun install --frozen-lockfile
+	cd web && bun run format:check
+	cd web && bun run lint:check
+	cd web && bun run build
+	cd web && bun run test:coverage
 
 # Git hooks — point git at the tracked .githooks/ directory.
 hooks:
--- a/README.md
+++ b/README.md
@@ -102,6 +102,25 @@ make hooks
 
 This points git at `.githooks/` (via `core.hooksPath`). On `git push`, the hook runs the same checks CI runs, scoped to whichever top-level area changed (`server/` or `web/`). Bypass with `PREPUSH_SKIP=1 git push` or `git push --no-verify` when truly needed.
 
+To run the full CI matrix manually (format + build + test + coverage on both halves):
+
+```bash
+make ci          # both halves
+make ci-server   # server only
+make ci-web      # web only
+```
+
+### Seeded test account
+
+`make seed` creates a known test user so contributors can sign in immediately on a fresh DB without configuring OAuth credentials:
+
+| Field    | Value                  |
+|----------|------------------------|
+| Email    | `seeduser@dev.local`   |
+| Password | `testpass123!`         |
+
+Use these credentials at `http://localhost:5173/login` (or `POST /auth/login` with `{ email, password }`).
+
 ## Docker
 
 ### Start Infrastructure
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -10,6 +10,15 @@ services:
       # Pin PGDATA to a subdir so the data path is stable across major-version
       # bumps (the upstream default changes per major) and avoids PG's refusal
       # to init into a non-empty mount root.
+      #
+      # MIGRATION NOTE: existing developers upgrading from the pre-PG18 layout
+      # have data at the volume root (`/var/lib/postgresql/data`), not the new
+      # `/pgdata` subdir — Postgres will treat the new path as empty and re-init
+      # a fresh cluster, leaving the old files orphaned. Local seed data is
+      # throwaway, so the canonical fix is `make clean && make up && make
+      # migrate && make seed`. If you have local data worth keeping, exec into
+      # the running container and `mv /var/lib/postgresql/data/{base,pg_*,*}
+      # /var/lib/postgresql/data/pgdata/` before restarting.
       PGDATA: /var/lib/postgresql/data/pgdata
     volumes:
       - postgres_data:/var/lib/postgresql/data
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,12 +1,16 @@
 services:
   postgres:
-    image: postgres:16
+    image: postgres:18
     ports:
       - "5435:5432"
     environment:
       POSTGRES_DB: gankedtv
       POSTGRES_USER: gankedtv
       POSTGRES_PASSWORD: gankedtv_dev
+      # Pin PGDATA to a subdir so the data path is stable across major-version
+      # bumps (the upstream default changes per major) and avoids PG's refusal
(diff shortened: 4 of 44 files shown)


Turbootzz:
@coderabbitai review

Turbootzz:
@coderabbitai review

Turbootzz:
@coderabbitai review

PR #79, opened 2026-05-15
feat: clip owner edit and delete actions

### LINKED ISSUE #71: [web] Clip owner actions: delete with confirm modal
## Context

`DELETE /clips/{id}` is wired server-side (owner-gated, S3 cleanup via `ClipBlobCleanup.TryDeleteAsync` — see `server/src/GankedTV.Api/Endpoints/ClipsMutateEndpoints.cs:27`). Web has no way to call it. Pair to #70 — same menu, same api module.

## Scope

- Add `clips.delete(id)` to `web/src/api/clips.ts`.
- Confirm modal component using `web/DESIGN.md` tokens. Keep it generic enough to reuse for future destructive actions (single `<ConfirmDialog>` taking title/body/confirmLabel slots).
- Wire the Delete entry in the same "more" menu as #70 (`web/src/views/ClipView.vue:270`); owner-only.
- On 204: redirect to `/me`, invalidate cached feed entries, show toast.

## Acceptance

- Owner can delete from clip page; non-owner cannot see the entry.
- After confirm, navigation lands on `/me` and the deleted clip is gone from the feed and the profile.
- Coverage stays ≥ 85% on scoped paths.

## Out of scope

- Bulk delete.
- Soft-delete / restore.

mintycw:
# Depends on #78  — review and merge that first.

## Summary
Adds a Delete action to the clip owner more-menu, gated behind a reusable confirmation modal. Also ships the clip edit dialog (wired to `PATCH /clips/{id}`) and a `GameSelector` component extracted from upload — the edit and delete actions ship together as the clip owner action set.

## What's here
- **ConfirmDialog.vue** — reusable destructive-confirm modal with danger variant, busy state, and accessible markup (`role=dialog`, `aria-modal`, `aria-labelledby`, autofocus on Cancel)
- **ClipEditDialog.vue** — owner-gated edit dialog for title/description/game/visibility via sparse-diff PATCH
- **GameSelector.vue** — debounced game search with popular chips, extracted from UploadView and shared between upload and edit
- **clips.ts** — `clips.delete(id)` wrapping `DELETE /clips/{id}` (204) and `clips.update()` for PATCH
- **Server** — adds `visibility` to `ClipDetailResponse` and fixes `GetDetail` to serve unlisted clips via direct link (feed gate only)

Closes #71

## How to test manually
1. Log in as a clip owner and open one of your clips.
2. Open the more-menu (⋯) — **Edit** and **Delete** appear; verify they're absent when viewing as another user.
3. **Delete flow:** click Delete → modal opens; Cancel closes it without a request; Confirm triggers DELETE, shows a toast, and redirects to your profile.
4. **Edit flow:** click Edit → dialog opens pre-filled; change a field and submit → clip updates in place.
5. Open an unlisted clip via direct link — it should load; confirm it doesn't appear in feeds.

## Checklist
- [ ] Verified manually

## New Features

* Clip owners can delete their clips from the clip view, accessible via the owner's menu. A confirmation dialog appears before deletion is completed, with support for canceling via escape key or clicking outside. Success and error notifications inform users of the deletion outcome. The dialog prevents interactions while the deletion request is processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/79?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/src/components/ClipEditDialog.vue
+++ b/web/src/components/ClipEditDialog.vue
@@ -45,9 +45,10 @@ const diff = computed((): UpdateClipBody => {
   const trimmedTitle = localTitle.value.trim()
   if (trimmedTitle !== props.clip.title) body.title = trimmedTitle
   // Normalize empty string ↔ null when comparing, but send "" to the server to clear.
-  const localDescNorm = localDesc.value === '' ? null : localDesc.value
+  const trimmedDesc = localDesc.value.trim()
+  const localDescNorm = trimmedDesc === '' ? null : trimmedDesc
   const originalDescNorm = props.clip.description || null
-  if (localDescNorm !== originalDescNorm) body.description = localDesc.value
+  if (localDescNorm !== originalDescNorm) body.description = trimmedDesc
   if ((localGame.value?.id ?? null) !== (props.clip.game?.id ?? null))
     body.gameId = localGame.value?.id ?? null
   if (localVisibility.value !== props.clip.visibility) body.visibility = localVisibility.value
--- a/web/src/components/ConfirmDialog.vue
+++ b/web/src/components/ConfirmDialog.vue
@@ -1,0 +1,114 @@
+<script setup lang="ts">
+import { ref, watch, nextTick, onMounted, onUnmounted } from 'vue'
+
+const props = defineProps<{
+  open: boolean
+  title: string
+  body: string
+  confirmLabel: string
+  cancelLabel?: string
+  variant?: 'danger' | 'default'
+  busy?: boolean
+}>()
+
+const emit = defineEmits<{
+  confirm: []
+  cancel: []
+}>()
+
+const cancelBtn = ref<HTMLButtonElement | null>(null)
+
+watch(
+  () => props.open,
+  (open) => {
+    if (open) nextTick(() => cancelBtn.value?.focus())
+  },
+)
+
+function onKeydown(e: KeyboardEvent) {
+  if (e.key === 'Escape' && props.open && !props.busy) emit('cancel')
+}
+
+onMounted(() => window.addEventListener('keydown', onKeydown))
+onUnmounted(() => window.removeEventListener('keydown', onKeydown))
+</script>
+
+<template>
+  <Teleport to="body">
+    <Transition
+      enter-active-class="transition-opacity duration-200"
+      enter-from-class="opacity-0"
+      leave-active-class="transition-opacity duration-150"
+      leave-to-class="opacity-0"
+    >
+      <div
+        v-if="open"
+        class="fixed inset-0 z-50 flex items-center justify-center px-4"
+        role="dialog"
+        aria-modal="true"
+        :aria-labelledby="`confirm-dialog-title-${title}`"
+      >
+        <!-- Backdrop -->
+        <div class="absolute inset-0 bg-black/70" @click="!busy && emit('cancel')" />
+
+        <!-- Dialog card -->
+        <div
+          class="relative z-10 w-full max-w-lg rounded-md border border-border bg-surface-raised shadow-[0_0_40px_var(--color-brand-glow)]"
+          @click.stop
+        >
+          <!-- Header -->
+          <div class="flex items-center justify-between border-b border-border px-5 py-4">
+            <h2
+              :id="`confirm-dialog-title-${title}`"
+              class="font-heading text-lg font-bold uppercase tracking-[0.04em] text-text-primary"
+            >
+              {{ title }}
+            </h2>
+            <button
+              type="button"
+              :disabled="busy"
+              @click="emit('cancel')"
+              aria-label="Close"
+              class="cursor-pointer font-mono text-xl leading-none text-text-muted transition-colors duration-150 hover:text-text-primary disabled:pointer-events-none disabled:opacity-40"
+            >
+              ×
+            </button>
+          </div>
+
+          <!-- Body -->
+          <div class="px-5 py-4 font-body text-sm leading-relaxed text-text-secondary">
+            {{ body }}
+          </div>
+
+          <!-- Footer -->
+          <div class="flex justify-end gap-2.5 border-t border-border px-5 py-4">
+            <button
+              ref="cancelBtn"
+              type="button"
+              :disabled="busy"
+              @click="emit('cancel')"
+              class="cursor-pointer rounded-md border border-border bg-surface-overlay px-5 py-2.5 font-heading text-sm font-bold uppercase tracking-wider text-text-secondary transition-colors duration-150 hover:text-text-primary disabled:pointer-events-none disabled:opacity-40"
+            >
+              {{ cancelLabel ?? 'Cancel' }}
+            </button>
+            <button
+              type="button"
+              :disabled="busy"
+              @click="emit('confirm')"
+              :class="[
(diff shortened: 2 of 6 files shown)


web/src/components/ConfirmDialog.vue
Turbootzz, line 1: ConfirmDialog builds invalid HTML IDs from the title.

With title="Delete clip?" the rendered id is confirm-dialog-title-Delete clip? HTML IDs cannot contain whitespace, and aria-labelledby lookup will not match cleanly (and any second instance on the same page would collide). Use useId() (Vue 3.5+) or a stable crypto.randomUUID() per instance instead of interpolating the title.

web/src/views/ClipView.vue
Turbootzz, line 1: Delete error toast surfaces raw server codes to the user.

```
const message = err instanceof ApiError
  ? ((err.body as { code?: string } | null)?.code ?? 'Failed to delete clip')
  : 'Failed to delete clip'
```
This will display strings like CLIP_NOT_OWNER to end users. ClipEditDialog already has the right pattern with an ERROR_CODES map — mirror that here (e.g., forbidden → "You don't have permission to delete this clip", not_found → "Clip not found"), or fall through to the generic message.

web/src/components/ClipEditDialog.vue
```diff
@@ -0,0 +1,237 @@
...
+
+// Build a sparse payload containing only changed fields.
+const diff = computed((): UpdateClipBody => {
+  const body: UpdateClipBody = {}
+  const trimmedTitle = localTitle.value.trim()
+  if (trimmedTitle !== props.clip.title) body.title = trimmedTitle
+  // Normalize empty string ↔ null when comparing, but send "" to the server to clear.
+  const localDescNorm = localDesc.value === '' ? null : localDesc.value
+  const originalDescNorm = props.clip.description || null
+  if (localDescNorm !== originalDescNorm) body.description = localDesc.value
```
Turbootzz: The Tile is being trimmed but the description is not, consider trimming the description aswell before sending

web/src/components/GameSelector.vue
```diff
@@ -0,0 +1,132 @@
+<script setup lang="ts">
+import { ref, watch, onMounted, onUnmounted } from 'vue'
+import { games as gamesApi, type GameListItem } from '@/api/games'
+import type { GameSummary } from '@/api/clips'
+import GameChipButton from '@/components/GameChipButton.vue'
+import GameSearchResult from '@/components/GameSearchResult.vue'
+
+defineProps<{ modelValue: GameSummary | null }>()
+const emit = defineEmits<{ 'update:modelValue': [value: GameListItem | null] }>()
```
Turbootzz: Structurally fine (GameListItem is a superset), but the parent ref typed as GameSummary silently gains an extra coverUrl field after a pick. Pick one type and use it both directions — or strip the value to GameSummary shape in pickGame.

server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
```diff
@@ -136,11 +136,13 @@ private static async Task<IResult> GetDetail(
         IOptions<S3Options> s3,
         CancellationToken ct)
     {
+        // Unlisted clips are accessible to anyone with the link — only the feed is gated
+        // to public-only. Visibility is enforced at the listing layer, not the detail layer.
         var clip = await db.Clips.AsNoTracking()
```
Turbootzz, line 141: Visibility policy change deserves an explicit OpenAPI/contract note.

flips GET /clips/{id} from "public-only" to "any visibility, link is the gate". The inline comment is good; the renamed test covers it. Worth checking the OpenAPI generator output mentions visibility is now part of

Turbootzz (commented):
Global keydown handlers stack. Both ClipEditDialog and ConfirmDialog install window.addEventListener('keydown', ...) on mount unconditionally. If the delete confirm modal opens on top of an open edit dialog (not currently possible per UX, but no guard), Escape would fire both handlers. Each handler does check props.open before acting, so the behavior is safe — but worth gating the listener install on open if more dialogs land.

and

We are missing test coverage for ClipEditDialog and ConfirmDialog and GameSelector. since they belong to our scoped src/api/* & src/router/* & src/stores/*. to keep coverage high we need tests for these

web/src/components/ConfirmDialog.vue
mintycw: Fixed in 8c53e7e — `useId()` (Vue 3.5+) replaces the interpolated title for both `id` and `aria-labelledby`.

web/src/views/ClipView.vue
mintycw: Fixed in 8c53e7e — added `DELETE_ERROR_CODES` map (forbidden / not_found / unauthorized) mirroring the ClipEditDialog pattern.

web/src/components/ClipEditDialog.vue
mintycw: Fixed in 8c53e7e — description is now trimmed before diff/PATCH, matching the title path.

web/src/components/GameSelector.vue
mintycw: Fixed in 8c53e7e — `pickGame` now strips to `GameSummary` shape (id/name/slug/tag) and the emit type matches the prop type both directions.

server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
mintycw: Verified — `ClipDetailResponse.Visibility` is a public record property and `Microsoft.AspNetCore.OpenApi` derives the schema from response types automatically (no checked-in generated artifact). Web `ClipDetail` already includes `visibility`, so the contract is consistent end-to-end.

web/src/components/ConfirmDialog.vue
```diff
@@ -0,0 +1,114 @@
...
+
+const props = defineProps<{
+  open: boolean
+  title: string
+  body: string
+  confirmLabel: string
+  cancelLabel?: string
+  variant?: 'danger' | 'default'
+  busy?: boolean
+}>()
```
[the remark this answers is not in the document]
mintycw: Skipping per YAGNI — `ConfirmDialog` has one caller (ClipView delete), passing plain strings. No current need for rich markup in title/body. Happy to revisit when a real caller needs HTML content.

web/src/components/ConfirmDialog.vue
```diff
@@ -0,0 +1,114 @@
...
+    >
+      <div
+        v-if="open"
+        class="fixed inset-0 z-50 flex items-center justify-center px-4"
+        role="dialog"
+        aria-modal="true"
+        :aria-labelledby="`confirm-dialog-title-${title}`"
+      >
+        <!-- Backdrop -->
+        <div class="absolute inset-0 bg-black/70" @click="!busy && emit('cancel')" />
```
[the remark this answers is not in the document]
mintycw: Skipping — `bg-black/70` is the existing backdrop convention in this repo (also used in `ClipEditDialog.vue:109`), and `text-white` on filled buttons is conventional contrast. Tokenizing only here would be inconsistent; this is a design-system pass, not a one-file fix.

Turbootzz (approved):

Issue #73, opened 2026-05-14
[server][web] Social embeds: OG/Twitter meta tags for /c/{code}


Turbootzz:
## Context

Phase 2's killer feature for virality — clip links must auto-embed in Discord, Twitter/X, Slack, Telegram, and similar with thumbnail and inline player. Plain SPA HTML has no usable OG tags, so crawlers see nothing.

**Depends on:** #72 (share_code column + resolver).

## Server scope

- New endpoint `GET /c/{code}` returning HTML (string templated — no Razor needed for ~30 lines):
  - **Social crawler UA** (`Discordbot`, `Twitterbot`, `facebookexternalhit`, `Slackbot`, `LinkedInBot`, `TelegramBot` — keep list in a static array): minimal HTML with `og:title`, `og:description`, `og:image` (thumbnail URL — presigned or public depending on the thumbnails bucket policy), `og:video`, `og:video:type=video/mp4`, `og:type=video.other`, `twitter:card=player`, `twitter:player`.
  - **Everyone else:** 302 redirect to the SPA route (`/clip/{id}` or `/c/{code}` once the web route lands).
- Resolve any conflict with the JSON route from #72 (Accept-header negotiation or move JSON to `/api/c/{code}`).
- Reuse the read helper extracted in #72 — do not duplicate clip lookup.

## Web scope

- Add `/c/:code` route in `web/src/router/index.ts`; resolve via the share-code API and render the existing `ClipView.vue`.
- Update the share button / copy-link affordance to emit the short URL.

## Acceptance

- Pasting a `/c/{code}` URL into Discord shows the clip with thumbnail and an inline player.
- Twitter card validator returns `player` for the URL.
- A normal browser visit lands on the SPA clip page.

## Out of scope

- SSR for human users (only crawlers get rendered HTML).
- oEmbed endpoint.
- og:video:width / height tuning — defer if metadata isn't yet stored.

PR #80, opened 2026-05-15
[server] Short share codes: add share_code column + GET /c/{code} resolver

### LINKED ISSUE #72: [server] Short share codes: add share_code column + /c/{code} resolver
## Context

Phase 2 calls for `ganked.tv/c/{shortcode}` share URLs (PLAN.md §9). Adding the column server-first keeps the DB authoritative and unblocks the OG-embed work (#73, depends on this).

## Scope

- New EF Core migration: add `share_code VARCHAR(12) UNIQUE NOT NULL` to `clips`. Backfill existing rows in the migration body with a generated code (use a deterministic seeded RNG so the migration is repeatable). The unique constraint already implies an index.
- Generation helper at `server/src/GankedTV.Api/Clips/ShareCodeGenerator.cs`: 8-char base62 (`[A-Za-z0-9]`), retry-on-collision (probability is negligible but handle the duplicate-key path).
- Generate on clip create — touch the existing create handler in `ClipsMutateEndpoints` only; do not change the upload/complete steps.
- New endpoint `GET /c/{code:length(6,12)}` returning the same DTO shape as `GET /clips/{id}`. Factor the read logic into a single private method or extension on `ClipsReadEndpoints` so both routes call the same code (DRY).
- Update `Contracts/Clips/ClipMappings.cs` so `ClipDetail` and `ClipFeedItem` include `share_code`.

## Routing conflict note

`/c/{code}` will eventually serve HTML for crawlers in #73. To avoid a collision now, choose **one** of:
- Mount this JSON route at `/api/c/{code}` and keep `/c/{code}` reserved for #73, **or**
- Mount at `/c/{code}` and content-negotiate later (#73 inspects `Accept` / `User-Agent`).

Document the choice in this issue before #73 starts.

## Acceptance

- New clips get a unique `share_code`.
- Migration backfills all existing rows; running the migration twice on a fresh DB succeeds.
- Both `/clips/{id}` and the new route return the same DTO content.
- Coverage ≥ 85% line/branch.

## Out of scope

- Web `/c/:code` route (covered by #73).
- HTML rendering / OG tags (#73).

mintycw:
## Summary

Closes the gap for `ganked.tv/c/{shortcode}` share URLs (Phase 2), unblocking OG embed work (#73). Every clip now carries a permanent 8-char base62 `share_code`; a new `GET /c/{code}` resolver returns the same response as `GET /clips/{id}`.

## What's here

- **Migration** `AddClipShareCode` — adds `share_code VARCHAR(12) UNIQUE NOT NULL`; existing rows backfilled deterministically with `substr(digest(id::text, 'sha256'), 'hex'), 1, 8)` via pgcrypto (repeatable, no seed drift)
- **`ShareCodeGenerator`** — 8-char base62, bias-free via `RandomNumberGenerator.GetInt32`, DB-collision retry up to 5 attempts; mirrors `UsernameGenerator` pattern
- **Clip create** — `ClipUploadService.CreateAsync` calls `GenerateUniqueAsync` before persisting; `shareCode` exposed in `ClipDetailResponse` and `ClipFeedItem`
- **`GET /c/{code:length(6,12)}`** — DRY with `GET /clips/{id}` via shared `ResolveClipByPredicateAsync` helper; same visibility/status filter, same 404 envelope

> **Routing decision for #73:** JSON resolver lives at `/c/{code}` (consistent with the repo's no-`/api`-prefix convention). #73 adds content negotiation to serve HTML to crawlers vs JSON to API callers.

Closes #72

## How to test manually

```bash
make up
cd server && dotnet ef database update --project src/GankedTV.Api

# check backfill
docker compose -f docker-compose.dev.yml exec postgres \
  psql -U gankedtv -d gankedtv -c "select id, share_code from clips limit 3;"

# start API in another terminal: make server
curl -s http://localhost:5050/c/<share_code> | jq '.shareCode,.id,.title'
```

## Checklist
- [x] 443/443 tests, 95.1% line / 85.9% branch coverage (gate: 85/85)
- [ ] Migration + endpoint smoke-tested locally

## Summary by CodeRabbit

* **New Features**
  * Clips now automatically receive unique, short alphanumeric share codes for easier sharing
  * Share codes are displayed in all clip details and feed responses
  * New endpoint available to retrieve clip information directly using a clip's share code, providing an alternative method to access clips

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/80?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
+++ b/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
@@ -1,10 +1,12 @@
 using System.Buffers.Text;
 using System.Globalization;
 using System.IdentityModel.Tokens.Jwt;
+using System.Linq.Expressions;
 using System.Security.Claims;
 using System.Text;
 using GankedTV.Api.Contracts.Clips;
 using GankedTV.Api.Data;
+using GankedTV.Api.Data.Entities;
 using GankedTV.Api.Problems;
 using GankedTV.Api.Services.ObjectStorage;
 using Microsoft.EntityFrameworkCore;
@@ -27,6 +29,7 @@ public static class ClipsReadEndpoints
         var group = app.MapGroup("/clips");
         group.MapGet("/feed", GetFeed);
         group.MapGet("/{id:guid}", GetDetail);
+        app.MapGet("/c/{code:length(6,12)}", GetByShareCode);
         return app;
     }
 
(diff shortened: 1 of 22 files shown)


server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
```diff
@@ -27,6 +29,7 @@ public static IEndpointRouteBuilder MapClipsReadEndpoints(this IEndpointRouteBui
         var group = app.MapGroup("/clips");
```
Turbootzz, line 29: Route registered at root, not on /clips group.
The new route bypasses any future middleware/filters added to the /clips group. Today there's nothing there, so it's harmless but later we might add something like rate-limiting or filters. so for consistancy WOULD be nice

mintycw: Intentional — `/c/{code}` is the *short* share-URL surface (think `gnk.tv/c/AbC12345`), so registering it at root is what makes it short in the first place; putting it under `/clips/c/{code}` would defeat the purpose. If we later add filters/rate-limiting that need to apply here, we can attach them to this specific route or share a tag with the `/clips` group rather than nesting under the prefix.

Turbootzz (approved):

PR #81, opened 2026-05-15
feat: social embeds (OG/Twitter meta tags) for /c/{code} share links

### LINKED ISSUE #73: [server][web] Social embeds: OG/Twitter meta tags for /c/{code}

mintycw:
# Depends on #80  — review and merge that first.

## Summary

Share links need to embed with thumbnail and inline player in Discord, Twitter/X, Slack, and Telegram — plain SPA links give crawlers nothing to work with. This PR makes `GET /c/{code}` serve crawler-specific OG/Twitter Card HTML while preserving the existing JSON and browser redirect paths.

Depends on #72 (share code column + short-URL resolver — already in this branch).

## What's here

- **Server — UA content negotiation on `GET /c/{code}`:** social crawlers receive OG/Twitter Card HTML with presigned thumbnail and video URLs; `Accept: application/json` returns the existing JSON DTO (SPA contract unchanged); everything else gets a `302` to `${WebOrigin}/c/{code}`
- **Server — `LoadClipWithUrlsAsync` extraction:** DRYs the DB lookup + URL signing shared by the JSON and HTML paths; all URL values in the HTML template are `HtmlEncode`d to handle `&` in presigned S3 query strings
- **Web — `/c/:code` route + types + API client:** `shareCode: string` added to `ClipFeedItem`/`ClipDetail`; `clips.getByShareCode()` added; `/c/:code` route registered (name: `clip-share`, reuses `ClipView`); `ClipView` generalised to load via either `clipId` or `shareCode` param; share button now copies the short URL

Closes #73.

## How to test manually

```bash
# Crawler gets OG HTML
curl -s -H 'User-Agent: Discordbot/2.0' http://localhost:5050/c/{code} | grep -E 'og:|twitter:'

# Browser gets 302 to SPA
curl -I -H 'User-Agent: Mozilla/5.0' http://localhost:5050/c/{code}
```

- Visit `http://localhost:5173/c/{code}` — clip loads and plays via the new `clip-share` route
- Click share on any clip — clipboard contains `http://localhost:5173/c/{code}` (short URL, not `/clip/{id}`)

## Checklist

- [ ] Verified manually (crawler UA → OG HTML, browser → redirect, SPA → plays, share button → short URL)

## Summary by CodeRabbit

* **New Features**
  * Shareable clip URLs (/c/:code) with client support and a new API call to load clip details by share code.
  * Social preview pages for crawlers (OG/Twitter metadata) and a canonical share action that copies the /c/<code> link.

* **Bug Fixes**
  * Preview HTML is properly HTML-escaped and uses the canonical web origin for og:url.

* **Tests**
  * Added integration and client tests covering share-code routing, negotiation, encoding, and crawler behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/81?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
+++ b/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
@@ -248,7 +248,9 @@ public static class ClipsReadEndpoints
     {
         var title = WebUtility.HtmlEncode(clip.Title);
         var desc = clip.Description is not null ? WebUtility.HtmlEncode(clip.Description) : null;
-        var canonicalUrl = $"{request.Scheme}://{request.Host}/c/{clip.ShareCode}";
+        var canonicalUrl = WebUtility.HtmlEncode($"{request.Scheme}://{request.Host}/c/{clip.ShareCode}");
+        var encodedVideoUrl = WebUtility.HtmlEncode(videoUrl);
+        var encodedThumbnailUrl = WebUtility.HtmlEncode(thumbnailUrl);
         // width/height fallback: Discord/Twitter require numeric values
         var width = clip.Width?.ToString() ?? "1280";
         var height = clip.Height?.ToString() ?? "720";
@@ -264,17 +266,17 @@ public static class ClipsReadEndpoints
             <meta property="og:url" content="{canonicalUrl}" />
             <meta property="og:title" content="{title}" />
             {(desc is not null ? $"""<meta property="og:description" content="{desc}" />""" : "")}
-            <meta property="og:image" content="{thumbnailUrl}" />
-            <meta property="og:video" content="{videoUrl}" />
-            <meta property="og:video:secure_url" content="{videoUrl}" />
+            <meta property="og:image" content="{encodedThumbnailUrl}" />
+            <meta property="og:video" content="{encodedVideoUrl}" />
+            <meta property="og:video:secure_url" content="{encodedVideoUrl}" />
             <meta property="og:video:type" content="video/mp4" />
             <meta property="og:video:width" content="{width}" />
             <meta property="og:video:height" content="{height}" />
             <meta name="twitter:card" content="player" />
-            <meta name="twitter:player" content="{videoUrl}" />
+            <meta name="twitter:player" content="{encodedVideoUrl}" />
             <meta name="twitter:player:width" content="{width}" />
             <meta name="twitter:player:height" content="{height}" />
-            <meta name="twitter:image" content="{thumbnailUrl}" />
+            <meta name="twitter:image" content="{encodedThumbnailUrl}" />
             </head>
             <body></body>
             </html>
--- a/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
+++ b/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
@@ -163,30 +163,41 @@ public static class ClipsReadEndpoints
 
         var (clip, videoUrl, thumbnailUrl) = result.Value;
         var userAgent = request.Headers.UserAgent.ToString();
+        var webOrigin = oauthOptions.Value.WebOrigin.TrimEnd('/');
 
+        // Crawler UA wins over Accept negotiation: a bot advertising application/json
+        // (rare but possible) still needs the OG HTML so previews render.
         if (IsCrawler(userAgent))
-            return Results.Content(BuildOgHtml(clip, videoUrl, thumbnailUrl, request), "text/html; charset=utf-8");
+            return Results.Content(BuildOgHtml(clip, videoUrl, thumbnailUrl, webOrigin), "text/html; charset=utf-8");
 
         if (request.Headers.Accept.ToString().Contains("application/json"))
-        {
-            var likedByMe = false;
-            if (TryGetUserId(principal, out var userId))
-            {
-                likedByMe = await db.Likes.AsNoTracking()
-                    .AnyAsync(l => l.ClipId == clip.Id && l.UserId == userId, ct);
-            }
-            var expiresAt = DateTimeOffset.UtcNow.Add(VideoUrlLifetime);
-            return Results.Ok(clip.ToDetail(videoUrl, expiresAt, thumbnailUrl, likedByMe));
-        }
+            return await BuildDetailResultAsync(clip, videoUrl, thumbnailUrl, principal, db, ct);
 
-        var webOrigin = oauthOptions.Value.WebOrigin.TrimEnd('/');
         return Results.Redirect($"{webOrigin}/c/{code}", permanent: false);
+    }
+
+    private static async Task<IResult> BuildDetailResultAsync(
+        Clip clip,
+        string videoUrl,
+        string thumbnailUrl,
+        ClaimsPrincipal principal,
+        GankedTvDbContext db,
+        CancellationToken ct)
+    {
+        var likedByMe = false;
+        if (TryGetUserId(principal, out var userId))
+        {
+            likedByMe = await db.Likes.AsNoTracking()
+                .AnyAsync(l => l.ClipId == clip.Id && l.UserId == userId, ct);
+        }
+        var expiresAt = DateTimeOffset.UtcNow.Add(VideoUrlLifetime);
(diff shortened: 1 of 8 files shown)


Turbootzz (commented):
Check this first. then approval yes

Issues / suggestions
1. Twitter player card likely won't render ([ClipsReadEndpoints.cs:278](vscode-webview://1cp45b0l3u6fo279ivmtc7r54duti7sbg145erbrhveq58duugq3/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs#L278))
twitter:player must point to an HTTPS iframe HTML page, not a raw .mp4, and Twitter requires application approval for player cards. Pointing it at the presigned mp4 will fall back to a summary-large-image at best. Consider either:

Drop twitter:card=player and use summary_large_image (Twitter will still show the thumbnail + title), or
Build a tiny /embed/{code} HTML page that wraps a <video> and use that as twitter:player.
Also missing: twitter:title / twitter:description, which Twitter does use when present.

2. Presigned URL lifetime vs. crawler cache mismatch
Crawlers (Discord especially) cache OG payloads aggressively. The 1-hour presigned video/thumbnail will outlive the crawler hit, but cached HTML served to users seeing the embed weeks later will reference expired URLs → broken poster + player. Not blocking, but worth a follow-up issue — a public CDN URL or longer-lived signed URL for thumbnails would mitigate.

3. Empty-string description renders an empty meta tag ([ClipsReadEndpoints.cs:255](vscode-webview://1cp45b0l3u6fo279ivmtc7r54duti7sbg145erbrhveq58duugq3/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs#L255))


var desc = clip.Description is not null ? WebUtility.HtmlEncode(clip.Description) : null;
If Description == "", this emits <meta property="og:description" content="" />. Use string.IsNullOrWhiteSpace(clip.Description) for the guard.

4. Canonical URL uses raw request.Scheme/request.Host ([ClipsReadEndpoints.cs:253](vscode-webview://1cp45b0l3u6fo279ivmtc7r54duti7sbg145erbrhveq58duugq3/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs#L253))
Behind a reverse proxy without UseForwardedHeaders, this produces the internal scheme/host (e.g. http://localhost:5050) in og:url and the redirect canonical. Prefer oauthOptions.Value.WebOrigin for the canonical URL — that's the user-facing origin and is already trusted config in the redirect path.

5. UA-only crawler detection is trivially spoofable
Not a real security concern (the same data is reachable via the JSON path), but worth a note: the substring list ([ClipsReadEndpoints.cs:240](vscode-webview://1cp45b0l3u6fo279ivmtc7r54duti7sbg145erbrhveq58duugq3/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs#L240)) is a static private static readonly string[] — good. Consider adding Mastodon, Pinterest, Iframely, Embedly if you care about more surfaces. bingbot/Googlebot are deliberately omitted (you don't want them indexing share URLs as canonical), so the current set is fine — just decide intentionally.

6. likedByMe path duplicates logic ([ClipsReadEndpoints.cs:163-172](vscode-webview://1cp45b0l3u6fo279ivmtc7r54duti7sbg145erbrhveq58duugq3/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs#L163-L172) vs [:222-228](vscode-webview://1cp45b0l3u6fo279ivmtc7r54duti7sbg145erbrhveq58duugq3/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs#L222-L228))
Both branches now do the same anon-or-userId likes lookup + ToDetail projection. Extract a small helper (BuildDetailResultAsync(clip, videoUrl, thumbnailUrl, principal, db, ct)) to keep the JSON paths in sync as the DTO evolves.

7. Minor: internal exposure of LoadClipWithUrlsAsync
It's marked internal but no test in the diff calls it directly. If unused outside this file, keep it private.

8. Missing test: Accept: application/json + non-crawler UA still gets JSON
Existing test sets Accept header; another covers no-Accept browser flow. The branch ordering is correct (crawler check first, then Accept, then redirect), so a crawler UA with Accept: application/json currently still returns HTML. That's probably the right call (crawlers win), but it's worth one assertion to lock that precedence in.

Risks
Low correctness risk server-side: existing JSON contract preserved, route constraint scopes the new behavior tightly.
Embed quality risk: Twitter player card is likely to fall back; Discord/Slack/Telegram/Facebook should work fine. Worth manually verifying each platform per your test checklist before claiming the issue closed.

mintycw:
Thanks for the reviews — addressed in 14303d1.

### @Turbootzz

| # | Item | Resolution |
|---|------|-----------|
| 1 | Twitter player card | Switched to `summary_large_image` + `twitter:title`/`twitter:description`. Agreed — `player` needs an HTTPS iframe page + Twitter app approval; raw `.mp4` falls back to summary anyway. Held off on a dedicated `/embed/{code}` page (YAGNI — `summary_large_image` is the right default until we want autoplay-in-Twitter specifically). |
| 2 | Presigned URL ↔ crawler cache lifetime | Noted, deferring per your "not blocking" call. Will open a follow-up issue. |
| 3 | Empty-string description → empty `<meta>` | Guard now uses `string.IsNullOrWhiteSpace`. New test locks both branches. |
| 4 | Canonical URL uses raw `request.Scheme/Host` | `BuildOgHtml` now takes `webOrigin` from config — same source the redirect path already trusted. New test asserts `og:url` uses `WebOrigin`. |
| 5 | UA-only crawler detection / more bot UAs | Keeping current set as you suggested; left intentional. |
| 6 | `likedByMe` duplication | Extracted `BuildDetailResultAsync` — both `GetByShareCode` (JSON branch) and `ResolveClipByPredicateAsync` go through it. |
| 7 | `internal` on `LoadClipWithUrlsAsync` | Made `private` — verified no callers outside this file (unlike `BuildThumbnailUrl`/`LoadLikedClipIdsAsync` which legitimately cross files). |
| 8 | Test for precedence (crawler UA + `Accept: application/json` → HTML) | Added `ShareCodeResolve_CrawlerUA_BeatsAcceptJsonHeader`. |

### @coderabbitai

Both nitpicks are effectively already satisfied:
- **Explicit `Accept: application/json` on `getByShareCode`** — `buildHeaders` in [`web/src/api/client.ts:71`](web/src/api/client.ts#L71) already sets `Accept: application/json` on every `api()` call. Adding it per-call would be a no-op.
- **Test for Accept header + URL encoding** — already in [`web/src/api/__tests__/clips.spec.ts:129-148`](web/src/api/__tests__/clips.spec.ts#L129-L148) (commit 4c46c82, which landed before this review).

Turbootzz (approved):

PR #82, opened 2026-05-16
[server] Rate limit clip mutate endpoints

### LINKED ISSUE #74: [server] Rate limit clip mutate endpoints (POST/PATCH/DELETE /clips, likes)
## Context

`server/src/GankedTV.Api/Auth/AuthRateLimiting.cs` already wires `AddRateLimiter` and a `CredentialsPolicy` for `/auth/login` and `/auth/register`. Extending the same infrastructure to clip write endpoints is cheap and prevents trivial abuse before public launch.

## Scope

- Add a `ClipsWritePolicy` (fixed window, partitioned by `userId` when authenticated and by remote IP otherwise — same pattern as the credentials policy). Tune at ~30 writes / minute as a starting point.
- Apply via `.RequireRateLimiting("ClipsWritePolicy")` on:
  - `POST /clips`
  - `PATCH /clips/{id}`
  - `DELETE /clips/{id}`
  - `POST /clips/{id}/like`
  - `DELETE /clips/{id}/like`
- Tests: 31st request within the window returns 429 with the standard error envelope.

## Acceptance

- New policy is registered and applied to the listed endpoints.
- Existing auth policy and its tests unchanged.
- Coverage ≥ 85%.

## Out of scope

- Distributed rate limiting (single-instance only; revisit in Phase 4 with Redis).
- Tuning per-endpoint limits — start uniform and iterate from telemetry.

Turbootzz:
## Summary

Pre-launch hardening for clip writes. Adds a 30/min fixed-window rate limiter to the clip mutation endpoints (the same in-process pattern the auth flow already uses for `/auth/login` and `/auth/register`), so a compromised or scripted token can't trivially abuse `POST /clips`, `PATCH/DELETE /clips/{id}`, or the like/unlike pair before public launch.

## What's here

- New `ClipsWritePolicy` in `Clips/ClipsRateLimiting.cs` — fixed window, partitioned by JWT `sub` claim (per-user buckets), with defensive IP / `"unknown"` fallbacks for the case where a future change drops `RequireAuthorization()` from a `/clips` group.
- Attached at group level on `ClipsUploadEndpoints`, `ClipsMutateEndpoints`, and `LikesEndpoints` — covers the five endpoints in the issue plus the upload-flow companions (`upload-url`, `complete`) which are server-side writes in the same flow.
- A single global `OnRejected` handler emits the project's standard RFC 7807 envelope with `code = "rate_limited"`. Side-effect: `/auth/login` 429 responses now also carry this body instead of being empty — no test relied on the empty body, but flagging in case any frontend code did.
- Integration tests (`ClipsRateLimitTests`): 31st `POST /clips` returns 429 with the envelope; 31st like/unlike returns 429; two distinct users do not share a bucket.
- Unit tests (`ClipsRateLimitingTests`) for the partition-key fallbacks the HTTP path can't reach (`Sub` missing, no claims, no IP, empty `Sub` value).

Closes #74

## How to test manually

```bash
make up
make server   # in one shell

# in another, mint a dev JWT via POST /dev/token (or use OAuth), then:
for i in (seq 1 31)
  curl -s -o /dev/null -w "%{http_code}\n" \
    -H "Authorization: Bearer $TOKEN" \
    -X POST http://localhost:5050/clips \
    -H 'Content-Type: application/json' \
    -d '{"title":"smoke","visibility":"public"}'
end
# expect: 30 × 200, then 429

# inspect the rejection body
curl -i -H "Authorization: Bearer $TOKEN" \
  -X POST http://localhost:5050/clips \
  -H 'Content-Type: application/json' \
  -d '{"title":"x"}'
# expect: Content-Type: application/problem+json
#         body extension `code` = "rate_limited"

# wait 60s, fire one more — should be 200 again (window replenished)
```

## Checklist

- [ ] Verified manually with a real token against `make server`
- [ ] Confirm SPA renders a sensible message when a clip create/like is rejected with 429 (this is a previously-impossible status code for these endpoints)

## Summary by CodeRabbit

* **New Features**
  * Added rate limiting to clip write operations (create, update, delete) and like/unlike actions
  * Rate limits are enforced per-user when authenticated or per-IP when anonymous
  * Rate-limited responses return HTTP 429 with standardized error details

* **Tests**
  * Added comprehensive integration tests validating rate-limit behavior across endpoints and users

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/82?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs
+++ b/server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs
@@ -34,27 +34,32 @@ public static class ClipsRateLimiting
                 .ExecuteAsync(ctx.HttpContext));
 
         options.AddPolicy<string>(ClipsWritePolicy, ctx =>
-        {
-            // RequireAuthorization() on every /clips write group ensures HttpContext.User is
-            // populated before the limiter runs (UseAuthentication is wired before
-            // UseRateLimiter in Program.cs). We still fall through to IP / "unknown" so the
-            // partition function is total — a misconfiguration that drops the auth middleware
-            // must not collapse every caller into a single bucket and bypass the limit.
-            var sub = ctx.User.FindFirst(JwtRegisteredClaimNames.Sub)?.Value
-                ?? ctx.User.FindFirst(ClaimTypes.NameIdentifier)?.Value;
-            var key = !string.IsNullOrEmpty(sub)
-                ? $"u:{sub}"
-                : $"ip:{ctx.Connection.RemoteIpAddress?.ToString() ?? "unknown"}";
-
-            return RateLimitPartition.GetFixedWindowLimiter(key, _ => new FixedWindowRateLimiterOptions
+            RateLimitPartition.GetFixedWindowLimiter(ResolvePartitionKey(ctx), _ => new FixedWindowRateLimiterOptions
             {
                 AutoReplenishment = true,
                 PermitLimit = WritePermitLimit,
                 Window = WriteWindow,
                 QueueLimit = 0,
                 QueueProcessingOrder = QueueProcessingOrder.OldestFirst,
-            });
-        });
+            }));
         return options;
+    }
+
+    // Internal so unit tests can exercise the fallback branches that the HTTP-level integration
+    // tests can't reach (RequireAuthorization rejects pre-limiter, so the per-IP and
+    // NameIdentifier paths are unreachable through real requests).
+    //
+    // RequireAuthorization() on every /clips write group ensures HttpContext.User is populated
+    // before the limiter runs (UseAuthentication is wired before UseRateLimiter in Program.cs).
+    // The IP / "unknown" fallbacks remain so the partition function is total — a misconfiguration
+    // that drops the auth middleware must not collapse every caller into a single bucket and
+    // bypass the limit.
+    internal static string ResolvePartitionKey(HttpContext ctx)
+    {
+        var sub = ctx.User.FindFirst(JwtRegisteredClaimNames.Sub)?.Value
+            ?? ctx.User.FindFirst(ClaimTypes.NameIdentifier)?.Value;
+        return !string.IsNullOrEmpty(sub)
+            ? $"u:{sub}"
+            : $"ip:{ctx.Connection.RemoteIpAddress?.ToString() ?? "unknown"}";
     }
 }
--- a/server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs
+++ b/server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs
@@ -1,3 +1,4 @@
+using System.Globalization;
 using System.IdentityModel.Tokens.Jwt;
 using System.Security.Claims;
 using System.Threading.RateLimiting;
@@ -16,6 +17,16 @@ public static class ClipsRateLimiting
 
     // 30 writes per minute. Per-user when the caller is authenticated, per-IP otherwise.
     // Fixed window (not sliding) — same shape as the credentials policy, no extra state.
+    //
+    // The 30 covers the WHOLE clip-write surface — every endpoint that attaches this policy
+    // shares one bucket per user: POST /clips, POST /clips/{id}/upload-url, POST /clips/{id}/complete,
+    // PATCH /clips/{id}, DELETE /clips/{id}, POST /clips/{id}/like, DELETE /clips/{id}/like.
+    // A happy-path upload burns 3 (create + upload-url + complete), so the floor is ~10 clips/min
+    // per user before the bucket exhausts. Pinned by ClipsRateLimitTests.MixedWrites_ShareBucket.
+    //
+    // SCALING CAVEAT: this is in-process, per-instance state. Once the API runs on more than
+    // one pod the effective limit becomes 30 × pod_count per user. Phase 4 (issue #74 out-of-scope)
+    // moves to a Redis-backed limiter for cluster-wide enforcement.
     public const int WritePermitLimit = 30;
     public static readonly TimeSpan WriteWindow = TimeSpan.FromMinutes(1);
 
@@ -29,9 +40,19 @@ public static class ClipsRateLimiting
         // slot. Wiring it here means every limiter rejection (this policy AND the credentials
         // policy registered in AuthRateLimiting) share one RFC 7807 envelope with
         // `code = "rate_limited"`, matching every other 4xx body the API emits.
+        //
+        // Sets `Retry-After` (seconds) from the lease metadata when available so well-behaved
+        // clients (and the SPA) can back off instead of polling blind. RFC 6585 §4.
         options.OnRejected = static (ctx, _) =>
-            new ValueTask(ProblemResults.TooManyRequests(RateLimitedCode)
+        {
+            if (ctx.Lease.TryGetMetadata(MetadataName.RetryAfter, out TimeSpan retryAfter))
+            {
+                ctx.HttpContext.Response.Headers.RetryAfter =
+                    ((int)retryAfter.TotalSeconds).ToString(CultureInfo.InvariantCulture);
+            }
+            return new ValueTask(ProblemResults.TooManyRequests(RateLimitedCode)
                 .ExecuteAsync(ctx.HttpContext));
+        };
 
         options.AddPolicy<string>(ClipsWritePolicy, ctx =>
             RateLimitPartition.GetFixedWindowLimiter(ResolvePartitionKey(ctx), _ => new FixedWindowRateLimiterOptions
--- a/server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs
+++ b/server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs
@@ -1,0 +1,60 @@
+using System.IdentityModel.Tokens.Jwt;
+using System.Security.Claims;
+using System.Threading.RateLimiting;
+using GankedTV.Api.Problems;
+using Microsoft.AspNetCore.RateLimiting;
+
+namespace GankedTV.Api.Clips;
+
+// Rate-limit policy for clip mutation endpoints (POST /clips, PATCH/DELETE /clips/{id},
+// the upload-url and complete steps, and the like/unlike pair). Kept out of Program.cs
+// so the policy + partition logic stay inside the coverage denominator — Program.cs is
+// excluded per CLAUDE.md.
+public static class ClipsRateLimiting
+{
+    public const string ClipsWritePolicy = "clips-write";
+
+    // 30 writes per minute. Per-user when the caller is authenticated, per-IP otherwise.
+    // Fixed window (not sliding) — same shape as the credentials policy, no extra state.
+    public const int WritePermitLimit = 30;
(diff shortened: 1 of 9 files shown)


mintycw (commented):
Reviewed the rate-limiting changes. No Critical issues — implementation is clean, well-commented, and faithful to the plan. Partitioning logic is sound, tests are real (no over-mocking), and the architectural choices fit existing patterns.

**Verdict: Ready to merge with fixes.** The genuine gaps are test-coverage on contracts the PR is implicitly relying on (auth-login 429 envelope, cross-endpoint shared bucket) and a missing `Retry-After` header. Inline comments below.

### Strengths

- Clean separation in `Clips/ClipsRateLimiting.cs`; keeps `Program.cs` to pure DI wiring (CLAUDE.md compliant).
- Partition fallback chain (`sub` → `NameIdentifier` → IP → `"unknown"`) mirrors existing `TryGetUserId` semantics. Empty-`sub` handling correctly avoids a collision bucket and is pinned by `ResolvePartitionKey_TreatsEmptySubAsMissing`.
- The "two users don't share bucket" integration test directly disproves IP-partitioning by interleaving through `127.0.0.1`.
- `ProblemResults.TooManyRequests` follows the existing convention exactly.
- Per-test factory re-creation (`InitializeAsync`) gives each test its own in-process limiter state — solid isolation.

### Follow-ups (out of scope, file separately)

- **SPA-side 429 handling** — `web/src/views/LoginView.vue:71` and `RegisterView.vue:59` handle 429 for auth, but no 429 handling exists for clip create/like. Users hitting the limit will get a generic error.
- **Config-bind the limits** (`30/min` writes, `5/min` auth) via `IOptions<T>`, matching the existing pattern for S3/Jwt/Maintenance. Not for v1, but worth tracking — operators currently can't tune without a redeploy.
- **PR description manual-test checklist** is unticked; at minimum recommend a hand-curl of the create-clip path showing the 429 envelope verbatim before merge.

server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    public static RateLimiterOptions AddClipsWritePolicy(this RateLimiterOptions options)
+    {
+        // OnRejected is a single global handler on RateLimiterOptions — there's no per-policy
+        // slot. Wiring it here means every limiter rejection (this policy AND the credentials
+        // policy registered in AuthRateLimiting) share one RFC 7807 envelope with
+        // `code = "rate_limited"`, matching every other 4xx body the API emits.
+        options.OnRejected = static (ctx, _) =>
+            new ValueTask(ProblemResults.TooManyRequests(RateLimitedCode)
+                .ExecuteAsync(ctx.HttpContext));
```
mintycw, line 54: **Important: the PR's claimed side-effect on `/auth/login` 429 isn't pinned by any test.**

This global `OnRejected` handler also fires for `AuthRateLimiting`'s credentials policy (which sets `RejectionStatusCode = 429` but no `OnRejected`). The PR description explicitly notes that auth-login 429 responses now carry the standard envelope body instead of being empty — but `AuthEndpointsRateLimitTests.Login_Exceeding5Per60s_Returns429` only asserts the status code, not the body.

If a future refactor splits `OnRejected` per-policy or moves it elsewhere, the auth-login envelope contract silently regresses.

**Fix:** Extend the auth limit test with one extra assertion mirroring `ClipsRateLimitTests.cs:91-93`:
```csharp
response.Content.Headers.ContentType?.MediaType.Should().Be("application/problem+json");
body.GetProperty(ProblemResults.CodeKey).GetString().Should().Be(ClipsRateLimiting.RateLimitedCode);
```

Alternatively (cleaner): extract a shared `AddSharedRateLimitDefaults()` extension and have both policy registrations call it.

mintycw, line 54: **Important: no `Retry-After` header on 429 responses.**

RFC 6585 §4 recommends `Retry-After` on 429. ASP.NET Core's rate limiter exposes this via lease metadata, and fixed-window with `AutoReplenishment = true` does emit it.

Without it, SPA-side handling can't render "try again in 14s" and well-behaved API consumers can't back off intelligently. Easy to add, real UX value before public launch:

```csharp
options.OnRejected = static (ctx, _) =>
{
    if (ctx.Lease.TryGetMetadata(MetadataName.RetryAfter, out TimeSpan retryAfter))
    {
        ctx.HttpContext.Response.Headers.RetryAfter =
            ((int)retryAfter.TotalSeconds).ToString();
    }
    return new ValueTask(ProblemResults.TooManyRequests(RateLimitedCode)
        .ExecuteAsync(ctx.HttpContext));
};
```

server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs
```diff
@@ -0,0 +1,65 @@
...
+// so the policy + partition logic stay inside the coverage denominator — Program.cs is
+// excluded per CLAUDE.md.
+public static class ClipsRateLimiting
+{
+    public const string ClipsWritePolicy = "clips-write";
+
+    // 30 writes per minute. Per-user when the caller is authenticated, per-IP otherwise.
+    // Fixed window (not sliding) — same shape as the credentials policy, no extra state.
+    public const int WritePermitLimit = 30;
+    public static readonly TimeSpan WriteWindow = TimeSpan.FromMinutes(1);
```
mintycw, line 31: **Important: cross-policy bucket contention is not tested.**

All three groups (`ClipsUploadEndpoints`, `ClipsMutateEndpoints`, `LikesEndpoints`) share one policy name → one bucket per user across upload + mutate + likes. A user calling `POST /clips` 25× and then `POST /{id}/like` 6× in the same minute will 429 on the 31st request.

This is a defensible design ("30 writes/min total") but it's not documented anywhere user-facing and isn't tested. A developer adding a fourth `RequireRateLimiting(ClipsWritePolicy)` group could shift the semantic without anyone noticing.

**Fix:** Add one integration test that mixes `POST /clips` and `POST /{id}/like` for the same authenticated user, summing to `WritePermitLimit + 1`, and assert the last request 429s. This pins "shared bucket" as a deliberate contract.

While here, a comment next to `WritePermitLimit` clarifying "30 covers the whole write surface" would prevent future churn.

mintycw, line 31: **Minor: multi-instance scaling caveat is undocumented.**

Once GankedTV runs more than one API instance (k8s, autoscale), each pod has its own bucket — effective limit becomes `30 × pod_count` per user. The comment in `AuthRateLimiting.cs:25-32` calls out the `X-Forwarded-For` caveat; an equivalent multi-instance note here would help future operators.

server/src/GankedTV.Api/Endpoints/ClipsUploadEndpoints.cs
```diff
@@ -17,7 +18,9 @@ public static class ClipsUploadEndpoints
 
     public static IEndpointRouteBuilder MapClipsUploadEndpoints(this IEndpointRouteBuilder app)
     {
-        var group = app.MapGroup("/clips").RequireAuthorization();
+        var group = app.MapGroup("/clips")
+            .RequireAuthorization()
+            .RequireRateLimiting(ClipsRateLimiting.ClipsWritePolicy);
```
mintycw, line 23: **Important — confirm intent: `POST /clips/{id}/upload-url` and `/complete` are now rate-limited at 30/min.**

Issue #74 lists `POST /clips`, `PATCH`, `DELETE`, and like/unlike — it does not explicitly list `/upload-url` or `/complete`. Applying the limiter at group level swept those in as a deviation.

A happy-path upload uses 3 endpoints per clip (`POST /clips` → `POST /upload-url` → `POST /complete`), so 30/min ÷ 3 ≈ 10 clips/min — plenty for a human, but worth flagging:
- A power-user batch script uploading on a fresh window may hit the wall earlier than they'd predict from counting only `POST /clips`.
- A retry loop on `/complete` (e.g. ffmpeg thumbnail-job slowness causing client retries) burns the same shared bucket.

Probably intentional — just want to confirm. If so, a one-line comment in `ClipsRateLimiting.cs` next to `WritePermitLimit` explaining "30 covers the whole write surface: create + upload-url + complete + patch + delete + like/unlike" would prevent future churn.

server/tests/GankedTV.Api.Tests/Integration/Endpoints/ClipsRateLimitTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        // in addition to /clips POST.
+        for (var i = 0; i < ClipsRateLimiting.WritePermitLimit; i++)
+        {
+            var resp = (i % 2 == 0)
+                ? await client.PostAsync($"/clips/{clipId}/like", content: null)
+                : await client.DeleteAsync($"/clips/{clipId}/like");
+            resp.StatusCode.Should().Be(HttpStatusCode.OK, $"request #{i + 1} should be allowed");
+        }
+
+        var blocked = await client.PostAsync($"/clips/{clipId}/like", content: null);
```
mintycw, line 122: **Minor:** Alternating POST/DELETE starting on POST with `i % 2 == 0` produces 15 inserts + 15 deletes; after the 30 alternations the row state is "no like" (last DELETE removed it). The 31st `POST /{id}/like` is then a fresh insert, which is what you want — the limiter rejects before the unique-key constraint on `(user_id, clip_id)` is involved. The inline comment already calls this out; just confirming the ordering invariant is intentional.

server/tests/GankedTV.Api.Tests/Integration/Endpoints/ClipsRateLimitTests.cs
```diff
@@ -0,0 +1,141 @@
...
+        // Each accepted POST returns 200 with a draft clip. We don't care about the body —
+        // only that the limiter lets exactly PermitLimit requests through, then blocks.
+        for (var i = 0; i < ClipsRateLimiting.WritePermitLimit; i++)
+        {
+            var ok = await client.PostAsJsonAsync("/clips", new { title = $"clip-{i}" });
+            ok.StatusCode.Should().Be(HttpStatusCode.OK, $"request #{i + 1} should be allowed");
+        }
+
+        var blocked = await client.PostAsJsonAsync("/clips", new { title = "over-limit" });
+        blocked.StatusCode.Should().Be(HttpStatusCode.TooManyRequests);
```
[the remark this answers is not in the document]
Turbootzz: Skipping — verified the concern is theoretical for this codebase.

- The framework's `FixedWindowRateLimiterOptions` (.NET 10) has no `TimeProvider` slot. Adopting the suggestion requires either replacing the framework limiter with a custom `TimeProvider`-aware implementation, or exposing `WriteWindow` via `IOptions<T>` so tests can shrink it. Both add real complexity.
- Margin of safety: window is 60s; the 30-request loop completes in ~1s on local hardware (in-process Kestrel → Postgres insert per call). ~60× margin against window rollover.
- Precedent: `AuthEndpointsRateLimitTests` uses the identical real-clock loop pattern and has been stable across CI. If clock skew were a real issue the auth tests would have flaked first.

Disproportionate to a not-yet-observed flake. Will reconsider if a CI run actually trips it.

mintycw (approved):

Issue #90, opened 2026-05-17
[server] Backfill game cover art + metadata from IGDB


Turbootzz:
## Summary

The `games` table already reserves `cover_url TEXT` and `igdb_id INT` ([PLAN.md §5](PLAN.md)) but nothing populates them today. Seeded rows ship with both null, so the games catalog tiles and the new `/game/:slug` header (#83) fall back to text-only rendering. This issue wires up IGDB as the metadata source so covers actually show up.

Surfaces affected:
- [web/src/views/GamesView.vue](web/src/views/GamesView.vue) — tile rendering
- [web/src/views/GameView.vue](web/src/views/GameView.vue) — header cover background (already `v-if`-gated on `coverUrl`, just needs a non-null value)
- Anywhere `GameSummary` / `GameListItem` / `GameDetail` flow client-side

## Server

- New `IGameMetadataService` backed by the IGDB API (Twitch OAuth client-credentials grant + rate-limited HTTP client; IGDB caps at 4 req/s).
- Config: `IGDB_CLIENT_ID` / `IGDB_CLIENT_SECRET` env vars, surfaced via typed `IgdbOptions`. Document in `appsettings.Development.json` + production env list.
- Cover storage decision (recommend MinIO):
  - **Option A (recommended): mirror to MinIO** under `game-covers/{igdb_id}.jpg`, store the relative key in `cover_url` and serve through our own CDN — matches the clip/thumbnail pattern, stable URLs, no third-party hotlinking.
  - **Option B: store IGDB-hosted URL directly** — simpler, but URLs can rotate and we lose CDN consistency.
- Backfill mechanism:
  - One-shot console command or `IHostedService` worker that walks `games` where `igdb_id IS NULL`, resolves by name, writes `igdb_id` + `cover_url`.
  - Idempotent — safe to re-run.
- Ongoing sync hook (smaller scope, can split): when a clip is uploaded tagging a game with `igdb_id IS NULL`, enqueue a metadata resolution job.

## Web

No client changes required. The header in [GameView.vue](web/src/views/GameView.vue) already renders the cover when present, and tiles in [GamesView.vue](web/src/views/GamesView.vue) inherit the same `coverUrl` field. Once rows have values, both surfaces light up automatically.

## Dev / seed

Update [server seeding](server/src/GankedTV.Api/Data/) so the ~10 seeded games either:
- have `igdb_id` pre-populated (deterministic, no IGDB call on every fresh dev DB), or
[...]

PR #91, opened 2026-05-17
[server][web] Game detail page: GET /games/{slug} + /game/:slug view 

### LINKED ISSUE #83: [server][web] Game detail page: GET /games/{slug} + /game/:slug view
## Summary

Phase 3 (Discovery) — game detail surface. Today there's a `GET /games` list endpoint and a `GamesView.vue` that shows the catalog, but clicking a game has nowhere to go. This issue adds the per-game landing page so a user can see every clip for a game and discover content by title.

Per [PLAN.md §9](PLAN.md) Phase 3: *Game pages: `/game/{slug}` with all clips for that game*.

## Server

Extend [server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs](server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs):

- `GET /games/{slug}` — returns `GameDetail` (id, slug, name, coverUrl, clipCount). 404 if the game doesn't exist.
- `GET /games/{slug}/clips` — cursor-paginated list of `ClipFeedItem`s for the game, reusing the keyset pattern from `ClipsReadEndpoints.GetFeed` (Base64Url `(CreatedAt, Id)` cursor). Filters to `Visibility = "public" AND Status = "ready"`.

Contracts:
- New `Contracts/Games/GameDetail.cs`
- New `Contracts/Games/GameMappings.cs` (currently missing — only DTOs exist in `Contracts/Games/`)

## Web

- New route `/game/:slug` (name: `game-detail`) in [web/src/router/index.ts](web/src/router/index.ts)
- New `web/src/views/GameView.vue` — header with cover image + name + clip count, infinite-scroll clips grid (reuse `ClipCard.vue`)
- Add `games.getBySlug(slug)` and `games.clips(slug, { cursor, limit })` to [web/src/api/games.ts](web/src/api/games.ts)
- Make the game chip on `ClipCard.vue` link to `/game/:slug` (today it's not a link)

## How to test manually

```bash
make up && make server  # one shell
cd web && bun dev       # another shell

# Hit the new endpoints directly
curl -s http://localhost:5050/games/valorant | jq
curl -s 'http://localhost:5050/games/valorant/clips?limit=10' | jq

# In the browser
open http://localhost:5173/games          # click any game → routes to /game/:slug
open http://localhost:5173/game/valorant  # clips grid renders, scrolling paginates
```

## Checklist

- [ ] 404 returned for unknown slug
- [ ] Cursor pagination produces no duplicates at the boundary
- [ ] Empty state when the game has no ready clips
- [ ] Coverage gate (85% line / 85% branch) still passes

Milestone: Phase 3 — Discovery · Source: [PLAN.md §9](PLAN.md)

Turbootzz:
## Summary

Phase 3 (Discovery) — the per-game landing page so clicking a game in the catalog finally goes somewhere. Adds `/games/{slug}` + `/games/{slug}/clips` on the server and a `/game/:slug` view on the web with an infinite-scroll grid of that game's clips.

## What's here

**Server**
- `GET /games/{slug}` → `GameDetail` (id, name, slug, tag, coverUrl, clipCount). 404 on unknown slug. `clipCount` matches the same `visibility='public' AND status='ready'` filter the clip list uses, so the header number equals what the user can actually scroll through.
- `GET /games/{slug}/clips` → cursor-paginated `ClipFeedResponse` for that game. 404 on unknown slug (distinguishes "no game" from "empty page").
- Extracted `ClipsReadEndpoints.BuildFeedPageAsync` so the global feed and per-game feed share the same keyset pagination (composite `(CreatedAt, Id)` cursor, `limit+1` fetch, likedByMe lookup, thumbnail signing). Zero duplicated pagination logic.
- New `Contracts/Games/GameMappings.cs` holds `ToGameSummary` (moved from `ClipMappings.cs`) and `ToDetail(clipCount)`. Mappers now live next to their DTOs per PLAN.md §11.

**Web**
- New `/game/:slug` route → `GameView.vue`: cover-image header, infinite scroll via `IntersectionObserver` (pre-fetches ~400px before the sentinel enters the viewport), separate first-load / not-found / pagination-error states, and a `watch(slug)` so navigating between two game pages resets state correctly.
- `games.getBySlug` / `games.clips` added to the API client, reusing `ClipFeedPage`/`ClipFeedQuery` from `clips.ts`.
- Game chip on `ClipCard.vue` wraps in a `RouterLink` to `game-detail` (with `@click.stop` so it doesn't bubble to the card's clip-open click).
- `GamesView.vue` tiles now navigate to `/game/:slug` instead of filtering in place — the previous filter mechanic was an explicit placeholder until the per-game endpoint existed.

**Tests**
- 8 new server integration tests covering detail+count, zero-count, 404s on both routes, game/visibility filtering, cursor boundary with shared `CreatedAt`, empty page, and invalid-cursor fallback. Existing 37 `ClipsReadEndpointsTests` still pass — guards the helper refactor.
- New web tests for `games.getBySlug` / `games.clips` URL construction and the ClipCard chip-as-link behavior.

Closes #83

## How to test manually

```bash
make migrate                # share_code column is required for the feed query
make up && make server      # one terminal
cd web && bun dev           # another

# API smoke
curl -s http://localhost:5050/games/valorant | jq
curl -s 'http://localhost:5050/games/valorant/clips?limit=10' | jq
curl -i http://localhost:5050/games/does-not-exist 2>&1 | head -1  # 404

# Browser walkthrough
open http://localhost:5173/games          # tile click routes to /game/:slug
open http://localhost:5173/game/valorant  # header + grid; scroll past bottom loads next page
open http://localhost:5173/game/no-such   # 404 panel + "Back to games"
open http://localhost:5173/               # game chip on any clip card → /game/:slug
```

To exercise the IntersectionObserver-driven infinite scroll you need ≥21 clips on one game — the dev seed has fewer, so seed extras if needed.

## Checklist

- [ ] Migrations applied (`make migrate`)
- [ ] Verified game tile + game chip both route to `/game/:slug`
- [ ] Verified infinite scroll loads a second page (needs ≥21 clips)
- [ ] Verified unknown slug returns the 404 panel with "Back to games"

## Follow-up

#90 tracks IGDB metadata backfill so `coverUrl` actually gets populated — the header in this PR already renders covers when present, it just degrades to text today because the seeded games have null `cover_url`.

## Summary by CodeRabbit

* **New Features**
  * Added game detail pages accessible via game navigation, displaying clips with infinite scroll pagination
  * Converted game browsing: Game cards now link to dedicated detail views instead of in-feed filtering
  * Game tag overlays on clip cards now navigate to the game's detail page

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/91?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/web/src/views/GameView.vue
+++ b/web/src/views/GameView.vue
@@ -33,12 +33,24 @@ const paginationErrored = ref(false)
 const sentinel = ref<HTMLElement | null>(null)
 let observer: IntersectionObserver | null = null
 
+// Monotonic request token. Bumped at the start of every loadAll (i.e. on mount
+// and every slug change). Each in-flight fetch captures the token at entry and
+// bails when it resolves if a newer loadAll has superseded it — otherwise a
+// stale response for slug A would stomp state belonging to slug B after a fast
+// nav. Also gates loading.value clear-down so a stale finally doesn't flip
+// loading off mid-load for the current request.
+let requestId = 0
+
 async function loadGame() {
   const s = slug.value
   if (!s) return
+  const token = requestId
   try {
-    game.value = await games.getBySlug(s)
+    const result = await games.getBySlug(s)
+    if (token !== requestId) return
+    game.value = result
   } catch (err) {
+    if (token !== requestId) return
     if (err instanceof ApiError && err.status === 404) {
       notFound.value = true
     } else {
@@ -51,14 +63,17 @@ async function loadGame() {
 async function loadMore() {
   const s = slug.value
   if (!s || loading.value || reachedEnd.value || notFound.value) return
+  const token = requestId
   loading.value = true
   paginationErrored.value = false
   try {
     const page = await games.clips(s, { cursor: cursor.value, limit: 20 })
+    if (token !== requestId) return
     items.value.push(...page.items)
     cursor.value = page.nextCursor
     if (page.nextCursor === null) reachedEnd.value = true
   } catch (err) {
+    if (token !== requestId) return
     if (items.value.length === 0) {
       // First-page failure (game lookup may have succeeded). Treat 404 here as
       // the same "game does not exist" branch so we don't render a broken header.
@@ -72,7 +87,7 @@ async function loadMore() {
     }
     console.error('game-detail: load failed', err)
   } finally {
-    loading.value = false
+    if (token === requestId) loading.value = false
   }
 }
 
@@ -95,6 +110,7 @@ function detachObserver() {
 }
 
 async function loadAll() {
+  const token = ++requestId
   errored.value = false
   notFound.value = false
   paginationErrored.value = false
@@ -111,16 +127,20 @@ async function loadAll() {
     // a 500 on /games/{slug} can set errored=true while /games/{slug}/clips
     // succeeds and silently populates items behind the error panel.
     await loadGame()
+    if (token !== requestId) return
     if (notFound.value || errored.value) return
     await loadMore()
   } catch {
     // individual handlers already set the right flag
   } finally {
-    initialLoading.value = false
-    // Defer observer attachment to next tick so the sentinel is in the DOM
-    // (it's only rendered when game is loaded and !errored).
-    if (!notFound.value && !errored.value && !reachedEnd.value) {
-      requestAnimationFrame(attachObserver)
+    // Bare `return` in finally is unsafe (masks throws), so gate via if-block.
+    if (token === requestId) {
+      initialLoading.value = false
+      // Defer observer attachment to next tick so the sentinel is in the DOM
+      // (it's only rendered when game is loaded and !errored).
+      if (!notFound.value && !errored.value && !reachedEnd.value) {
+        requestAnimationFrame(attachObserver)
+      }
     }
   }
 }
--- a/web/src/views/GameView.vue
+++ b/web/src/views/GameView.vue
@@ -1,0 +1,245 @@
+<script setup lang="ts">
+import { computed, onBeforeUnmount, onMounted, ref, watch } from 'vue'
+import { useRoute, useRouter } from 'vue-router'
+import { games, type GameDetail } from '@/api/games'
+import type { ClipFeedItem } from '@/api/clips'
+import { ApiError } from '@/api/client'
+import ClipCard from '@/components/ClipCard.vue'
+import GameTag from '@/components/GameTag.vue'
+import PageHeader from '@/components/PageHeader.vue'
+import StatusPanel from '@/components/StatusPanel.vue'
+
+const route = useRoute()
+const router = useRouter()
+
+const slug = computed(() => {
+  const raw = route.params.slug
+  return Array.isArray(raw) ? raw[0] : raw
+})
+
+const game = ref<GameDetail | null>(null)
+const items = ref<ClipFeedItem[]>([])
+const cursor = ref<string | null>(null)
+// Hit-end flag: distinguishes "more pages exist" (cursor !== null) from "first
+// page returned an empty list but the game does exist". Needed so the observer
+// doesn't keep firing forever and so the empty-state copy renders correctly.
+const reachedEnd = ref(false)
+const loading = ref(false)
+const initialLoading = ref(true)
+const errored = ref(false)
+const notFound = ref(false)
+const paginationErrored = ref(false)
+
+const sentinel = ref<HTMLElement | null>(null)
+let observer: IntersectionObserver | null = null
+
+async function loadGame() {
+  const s = slug.value
+  if (!s) return
+  try {
+    game.value = await games.getBySlug(s)
+  } catch (err) {
+    if (err instanceof ApiError && err.status === 404) {
+      notFound.value = true
+    } else {
+      errored.value = true
+    }
+    throw err
+  }
+}
+
+async function loadMore() {
+  const s = slug.value
+  if (!s || loading.value || reachedEnd.value || notFound.value) return
+  loading.value = true
+  paginationErrored.value = false
+  try {
+    const page = await games.clips(s, { cursor: cursor.value, limit: 20 })
+    items.value.push(...page.items)
+    cursor.value = page.nextCursor
+    if (page.nextCursor === null) reachedEnd.value = true
+  } catch (err) {
+    if (items.value.length === 0) {
+      // First-page failure (game lookup may have succeeded). Treat 404 here as
+      // the same "game does not exist" branch so we don't render a broken header.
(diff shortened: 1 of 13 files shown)


web/src/views/GameView.vue
```diff
@@ -0,0 +1,265 @@
...
+
+    <template v-else-if="game">
+      <!-- Header -->
+      <section
+        class="relative mb-10 overflow-hidden rounded-lg border border-border bg-surface-raised"
+      >
+        <div
+          v-if="game.coverUrl"
+          class="absolute inset-0 bg-cover bg-center opacity-30"
+          :style="{ backgroundImage: `url(${game.coverUrl})` }"
```
mintycw: **Important: CSS-injection foot-gun via `coverUrl`.**

```
:style="{ backgroundImage: `url(${game.coverUrl})` }"
```

A server-provided string is interpolated directly into a CSS `url()` value. A malicious or corrupted `coverUrl` containing `")` followed by additional CSS would break out, e.g. `a.jpg"); background: red; content: url("`. Today the value comes from the DB and is admin-controlled, so practical risk is low — but the fix is trivial and removes the foot-gun permanently.

Options: (a) validate on the server that `coverUrl` is an `http(s)://` URL with no embedded quotes before persisting; (b) render the cover as an `` stacked absolutely instead of a `background-image`; (c) at minimum, refuse to render if the URL fails a simple `URL` constructor check on the client.

mintycw (commented):
Self-review notes from a code-review pass. No Critical issues; flagging Important items inline. Verdict: ready to merge once these are addressed.

**Strengths**
- `BuildFeedPageAsync` extraction is well-motivated; existing 37 feed tests guard the refactor.
- Single-roundtrip detail+count with aligned filtering so the header number matches scrollable reality.
- 404 vs empty-page distinction taken seriously in both API and UI, with tests on both sides.
- Monotonic `requestId` token correctly cancels stale fetches on rapid slug navigation.
- Cursor-boundary test with shared `CreatedAt` actually exercises the composite `(CreatedAt, Id)` keyset.
- ClipCard chip stops both pointer and keyboard activation with the right Enter-vs-Space distinction.

**Minor (non-blocking)**
- Commit `1bf776f` titled `request counter` is the substantive race-fix commit — consider rewording before merge for `git log` archaeology.
- `limit=20` is hardcoded on the web side and also as the server default; drop the literal to keep one source of truth.
- `loadGame` re-throwing purely to break out of `loadAll`'s try block would read clearer as an explicit success/failure return.
- Tests don't exercise `?limit=99999` clamping against the per-game route specifically (the shared helper does cover it).
- Side note (pre-existing, not introduced here): `using GankedTV.Api.Auth;` in `ClipsReadEndpoints.cs` appears unused — no `Auth/*.cs` symbols are referenced in that file. Worth cleaning up in an unrelated commit.

web/src/views/GameView.vue
```diff
@@ -0,0 +1,265 @@
...
+          to="/upload"
+          class="rounded-sm border border-border bg-surface-overlay px-4 py-2 font-mono text-xs uppercase tracking-widest text-text-primary"
+        >
+          Upload a clip
+        </RouterLink>
+      </StatusPanel>
+
+      <!-- Sentinel (only when there's more to load) -->
+      <div
+        v-if="!reachedEnd"
```
mintycw: **Important: infinite silent retry on persistent pagination failure (block spans lines 239-248).**

When `paginationErrored` is true the sentinel stays in the DOM (`v-if="!reachedEnd"`) and the observer stays attached. The Retry button works, but if the user scrolls the sentinel out of view and back in, the observer fires again and silently re-retries the failing request — exactly the wrong UX during a real outage (the client will hammer the API on every micro-scroll).

Fix: when `paginationErrored` is set, also `detachObserver()`; re-attach inside the explicit Retry handler. Alternatively gate intersection-driven `loadMore` calls behind `!paginationErrored`.

web/src/views/GameView.vue
```diff
@@ -0,0 +1,265 @@
...
+        ref="sentinel"
+        class="mt-8 flex items-center justify-center py-6"
+        aria-hidden="true"
+      >
+        <span v-if="loading" class="font-mono text-[11px] uppercase tracking-widest text-text-muted"
```
mintycw: **Important: visible "Loading more…" text inside `aria-hidden="true"` container (lines 239-248).**

The sentinel `<div>` carries `aria-hidden="true"` (correct for the sentinel itself), but the `<span v-if="loading">Loading more…</span>` lives inside it and is therefore announced to nobody. Screen-reader users get no signal that more content is loading.

Fix: move the loading message into a sibling element with `role="status" aria-live="polite"`, leaving only the empty sentinel hidden.

web/src/views/GameView.vue
```diff
@@ -0,0 +1,265 @@
...
+          :key="clip.id"
+          :clip="clip"
+          @click="router.push({ name: 'clip', params: { id: clip.id } })"
+        />
+      </div>
+
+      <!-- Empty -->
+      <StatusPanel v-else-if="reachedEnd" kind="empty" message="No clips for this game yet.">
+        <RouterLink
+          to="/upload"
```
mintycw, line 257: **Important: empty-state CTA assumes the user can upload.**

The "No clips for this game yet" CTA is a `RouterLink to="/upload"`. `/upload` requires auth (router `beforeEach`), so an unauthenticated visitor clicking it bounces to login with a redirect — functional, but the CTA leads somewhere they can't go. Consider gating on `auth.isAuthenticated`, or showing a sign-in prompt instead.

web/src/views/GameView.vue
```diff
@@ -0,0 +1,265 @@
...
+    } else {
+      errored.value = true
+    }
+    throw err
+  }
+}
+
+async function loadMore() {
+  const s = slug.value
+  if (!s || loading.value || reachedEnd.value || notFound.value) return
```
mintycw, line 67: **Minor: `loading.value` can wedge true across a slug change.**

Trace: `loadAll` runs for slug A, `loadMore` starts (token = 1, `loading.value = true`). User navigates to slug B: `loadAll` runs, `++requestId` makes it 2, then calls `loadMore` — which early-returns here because `loading.value` is still `true`. The stale fetch eventually resolves, but the `finally` (line 90) only clears `loading.value` if `token === requestId` (1 !== 2), so it stays `true` forever for slug B and no further `loadMore` from intersection events will fire either.

Fix: clear `loading.value = false` in the `loadAll` reset block (alongside `cursor.value = null`, `items.value = []`).

mintycw (approved):

PR #111, opened 2026-05-23
discord.js implementation

### LINKED ISSUE #107: [server] Discord bot: auto-post clips to subscribed servers
## Summary

Phase 4 (Scale & Grow) — a Discord bot that auto-posts new clips to subscribed servers/channels. Clip share links already render rich embeds in Discord thanks to the OG/Twitter meta tags shipped in #73/#81, so a posted link unfurls into a thumbnail + title automatically. This is **huge for virality in gaming communities** ([PLAN.md §11](PLAN.md)).

Per [PLAN.md §9](PLAN.md) Phase 4: *Discord bot for auto-posting clips to servers*.

> **Largest standalone item in Phase 4** — likely to be broken into sub-issues during grooming (bot scaffolding, subscription model, posting pipeline). Captured here as the umbrella.

Relies on existing OG embeds (#73/#81). No other hard dependency.

## Scope

- **Bot/worker:** a standalone service — either a new `server/src/GankedTV.Bot` project or a hosted service in the API (off by default, like `IgdbSyncHostedService`). On each new `public` + `ready` clip (event hook on completion, or polling `GET /clips/feed`), post the share URL (`/c/{shareCode}`) to subscribed channels; Discord unfurls the embed.
- **Subscription model:** `discord_subscriptions` table (guild id, channel id, optional filter by game and/or creator), plus management — slash commands in-Discord and/or API endpoints. De-dupe so a clip posts once per channel.
- **Config/deploy:** `DISCORD_BOT_TOKEN` env, compose service, **off by default**; no-op without a token.

## How to test manually

- Invite the bot to a test guild, subscribe a channel (optionally filtered to a game).
- Upload/import a public clip → within the poll/post interval, the bot posts the `/c/{code}` link and Discord renders the embed.
- Unsubscribe → no further posts. Restart with no token → bot no-ops.

## Checklist

- [ ] Subscribing a channel + a new public clip → exactly one message posted, embed renders
- [ ] Game/creator filter respected
- [ ] No duplicate posts for the same clip in the same channel
- [ ] Bot no-ops when `DISCORD_BOT_TOKEN` is unset
- [ ] Unsubscribe stops posts

Milestone: Phase 4 — Scale & Grow · Source: [PLAN.md §9](PLAN.md) · Related: #73, #81 (OG embeds)

Turbootzz:
## Summary

Implements the Phase 4 Discord bot from #107 as a sibling `discord/` package (TypeScript + Bun + discord.js 14.26). The bot polls `/clips/feed` every 30s, fans new public+ready clips out to subscribed Discord channels, and exposes slash commands for subscription management and on-demand clip pulls. Off by default — boots and exits cleanly without `DISCORD_BOT_TOKEN`, matching the `IgdbSyncHostedService` contract.

## What's here

- **New `discord/` package** ([discord/src/](discord/src/)) — boot + config + db layer + poller + per-channel fanout + slash commands. 99.81% line / 98.72% function coverage, gate enforced by [discord/scripts/check-coverage.ts](discord/scripts/check-coverage.ts) (Bun's `coverageThreshold` is documentation-only on 1.3.13, so we parse `coverage/lcov.info` ourselves).
- **Three bot-owned tables** in the shared Postgres ([discord/src/migrations/001_initial.sql](discord/src/migrations/001_initial.sql)): `discord_subscriptions`, `discord_post_log`, `discord_bot_state`. EF Core doesn't model them — bot applies its own SQL migrations on boot via [discord/src/migrator.ts](discord/src/migrator.ts). `UNIQUE NULLS NOT DISTINCT` on the subscriptions key prevents phantom firehose duplicates.
- **Polling-based detection** ([discord/src/poller.ts](discord/src/poller.ts)) — high-water-mark cursor by `createdAt`, walks `nextCursor` to absorb bursts > `pageSize`, per-`(channel, clip)` post-log row as the at-least-once dedupe guard (post first, record on success). Future upgrade to push-based delivery only swaps the poller's entry-point.
- **Slash commands**: `/gankedtv subscribe|unsubscribe|subscriptions|pause|resume` (channel-scoped, gated to `ManageChannels`) and `/clip latest|top|search` (open in any guild, even without a subscription). Game autocomplete returns slugs uniformly across both namespaces.
- **Compose service** ([docker-compose.dev.yml](docker-compose.dev.yml)) — gated behind the `discord` profile so `make up` still spins only Postgres + MinIO. `extra_hosts: host.docker.internal:host-gateway` keeps the container reachable from the host-running API on native Linux Docker.
- **Repo wiring**: [Makefile](Makefile) (`make discord`, `make discord-install`, `make discord-test`, `make discord-lint`, `make ci-discord`, threaded into `make ci`), [.githooks/pre-push](.githooks/pre-push) (third scoped subset for `discord/` changes), [.github/workflows/discord.yml](.github/workflows/discord.yml) (build → format/lint → test+coverage with markdown summary in `$GITHUB_STEP_SUMMARY`), [scripts/new-worktree.sh](scripts/new-worktree.sh) (`make discord-install` + per-worktree `DISCORD_DATABASE_URL` in libpq form).
- **Env loading** ([discord/src/loadEnv.ts](discord/src/loadEnv.ts)) — reads the repo-root `.env` first, then layers `discord/.env`, then shell env (first-set-wins). Mirrors web's `envDir: '../'` convention. `DISCORD_DATABASE_URL` is a distinct env var from the API's `DATABASE_URL` because the API stores its connection string in dotnet semicolon form, which porsager's `postgres` driver can't parse.

Closes #107

## How to test manually

Per the original issue's manual-test checklist:

1. `make up && make migrate && make seed` to get the dev stack running.
2. Set `DISCORD_BOT_TOKEN`, `DISCORD_BOT_APP_ID`, and `DISCORD_BOT_GUILD_ID` in your root `.env` (the bot/app pair from https://discord.com/developers/applications; the guild id makes slash-command registration instant). Add `DISCORD_DATABASE_URL=postgres://gankedtv:gankedtv_dev@localhost:5435/gankedtv`.
3. `make discord` to start the bot (or `cd discord && bun run dev`). Should log `Discord client ready` and `Starting poller`.
4. Invite the bot to a test guild, run `/gankedtv subscribe` in a channel.
5. Upload a public clip via the web app (`make web`) → within one poll interval (30s default) the bot posts `https://gankedtv.com/c/<code>` and Discord unfurls the OG embed.
6. Subscribe with `game=valorant`; upload a Valorant clip + a non-Valorant clip; confirm only the Valorant one posts.
7. Restart the bot mid-poll → previously-logged clips don't double-post (post-log dedupe guard).
8. `/gankedtv unsubscribe` → no further posts.
9. Run `/clip latest`, `/clip top window:7d`, `/clip search query:flick` → bot posts the corresponding share URL.
10. Restart with `DISCORD_BOT_TOKEN` unset → logs `Discord bot disabled (...); exiting` and the process exits cleanly.

Also confirms the `enhancement` checklist from #107:

- [x] Subscribing a channel + a new public clip → exactly one message posted, embed renders
- [x] Game/creator filter respected
- [x] No duplicate posts for the same clip in the same channel
- [x] Bot no-ops when `DISCORD_BOT_TOKEN` is unset
- [x] Unsubscribe stops posts

## Checklist

- [x] Verified manually against a real Discord guild
- [x] Smoke-tested boot with no token (returns "disabled; exiting" and exits 0)
- [x] `make ci-discord` green locally

## Release Notes

* **New Features**
  * Added Discord bot with slash commands (`/gankedtv`, `/clip`) for subscribing to and viewing clips
  * Automated clip feed polling with filtering by game and creator
  * Optional role mention notifications on new clip posts
  * Game autocomplete in subscription commands

* **Configuration**
  * Added Discord bot environment variables and Docker profile support
  * New development commands for Discord bot setup and testing

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/111?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/discord/scripts/check-coverage.ts
+++ b/discord/scripts/check-coverage.ts
@@ -1,0 +1,105 @@
+#!/usr/bin/env bun
+// Enforces the discord/ coverage gate by parsing coverage/lcov.info and failing
+// when totals fall below thresholds. Bun's own `coverageThreshold` setting in
+// bunfig.toml prints the report but does NOT exit non-zero (verified on 1.3.13),
+// so without this script `make ci-discord` would silently pass under-covered
+// changes. Vitest enforces natively (see web/vitest.config.ts thresholds), so
+// this script exists only because Bun's enforcement is broken — once that's
+// fixed upstream, the script can be deleted and the bunfig.toml threshold
+// becomes the source of truth.
+//
+// Thresholds match web/: 85% lines, 85% functions (Bun's coverage doesn't emit
+// branch data, so we gate on functions instead — equivalent coverage signal in
+// practice for code that uses small focused functions).
+
+import { appendFile, readFile } from 'node:fs/promises';
+import { existsSync } from 'node:fs';
+
+const LCOV_PATH = 'coverage/lcov.info';
+const LINE_THRESHOLD = 0.85;
+const FUNCTION_THRESHOLD = 0.85;
+
+// Files included in test discovery but excluded from the coverage denominator —
+// kept in sync with bunfig.toml's coveragePathIgnorePatterns. Listed by prefix
+// match against the SF: path emitted by bun's lcov reporter.
+const EXCLUDE_PREFIXES = [
+  'src/index.ts',
+  'src/migrations/',
+  'src/migrator.ts',
+  'src/db.ts',
+  'src/api.ts',
+  'tests/',
+];
+
+if (!existsSync(LCOV_PATH)) {
+  console.error(`coverage gate: ${LCOV_PATH} not found — did 'bun test --coverage' run?`);
+  process.exit(1);
+}
+
+const lcov = await readFile(LCOV_PATH, 'utf8');
+
+let lineFound = 0;
+let lineHit = 0;
+let fnFound = 0;
+let fnHit = 0;
+let currentFile = '';
+let includeCurrent = true;
+
+for (const line of lcov.split('\n')) {
+  if (line.startsWith('SF:')) {
+    currentFile = line.slice(3);
+    includeCurrent = !EXCLUDE_PREFIXES.some((p) => currentFile.startsWith(p));
+    continue;
+  }
+  if (!includeCurrent) continue;
+  if (line.startsWith('LF:')) lineFound += Number(line.slice(3));
+  else if (line.startsWith('LH:')) lineHit += Number(line.slice(3));
+  else if (line.startsWith('FNF:')) fnFound += Number(line.slice(4));
+  else if (line.startsWith('FNH:')) fnHit += Number(line.slice(4));
+}
+
+const linePct = lineFound === 0 ? 1 : lineHit / lineFound;
+const fnPct = fnFound === 0 ? 1 : fnHit / fnFound;
+
+const fmt = (p: number) => (p * 100).toFixed(2) + '%';
+const linesLabel = `${fmt(linePct)} (${lineHit}/${lineFound})`;
+const fnLabel = `${fmt(fnPct)} (${fnHit}/${fnFound})`;
+const thresh = (n: number) => `${(n * 100).toFixed(0)}%`;
+
+console.log('');
+console.log('Discord coverage gate:');
+console.log(`  Lines:     ${linesLabel}  (threshold ${thresh(LINE_THRESHOLD)})`);
+console.log(`  Functions: ${fnLabel}  (threshold ${thresh(FUNCTION_THRESHOLD)})`);
+
+let failed = false;
+if (linePct < LINE_THRESHOLD) {
+  console.error(`  ✗ line coverage ${fmt(linePct)} below ${thresh(LINE_THRESHOLD)}`);
+  failed = true;
+}
+if (fnPct < FUNCTION_THRESHOLD) {
+  console.error(`  ✗ function coverage ${fmt(fnPct)} below ${thresh(FUNCTION_THRESHOLD)}`);
+  failed = true;
+}
+
+// Emit a GitHub Actions step summary (markdown table) when the runner exposes
+// the file. Mirrors web.yml's coverage-summary step but written in TS instead
+// of a shell+jq snippet to keep the parsing in one place.
+const summaryPath = process.env.GITHUB_STEP_SUMMARY;
+if (summaryPath) {
+  const status = failed ? '❌ failed' : '✅ passed';
+  const md = [
+    `## Discord coverage (scoped: src/, excludes I/O wrappers + boot)`,
+    ``,
+    `Gate: ${status}`,
+    ``,
+    `| Metric    | %     | Covered / Total | Threshold |`,
+    `|-----------|-------|-----------------|-----------|`,
+    `| Lines     | ${fmt(linePct)} | ${lineHit}/${lineFound} | ${thresh(LINE_THRESHOLD)} |`,
+    `| Functions | ${fmt(fnPct)} | ${fnHit}/${fnFound} | ${thresh(FUNCTION_THRESHOLD)} |`,
+    ``,
+  ].join('\n');
+  await appendFile(summaryPath, md);
+}
+
+if (failed) process.exit(1);
+console.log('  ✓ coverage gate passed');
--- a/discord/scripts/check-coverage.ts
+++ b/discord/scripts/check-coverage.ts
@@ -22,12 +22,29 @@ const FUNCTION_THRESHOLD = 0.85;
 // Files included in test discovery but excluded from the coverage denominator —
 // kept in sync with bunfig.toml's coveragePathIgnorePatterns. Listed by prefix
 // match against the SF: path emitted by bun's lcov reporter.
+//
+// What's NOT here on purpose:
+//   - src/api.ts — pure URL building + fetch + JSON parse, mockable via the
+//     injected fetchImpl param (tests/api.test.ts). Reviewer M1 from PR #111.
+//   - src/poller.ts, commands/**, filters.ts, lib/**, loadEnv.ts, config.ts,
+//     posting.ts, replies.ts — all testable with pure-function or fake-injection
+//     patterns; live in the denominator.
+//
+// What IS excluded (deliberately):
+//   - src/index.ts — boot wiring (Discord client login, signal handlers,
+//     compose-time env reads). Covered indirectly when the bot boots.
+//   - src/db.ts — postgres-js queries that need a real Postgres to validate
+//     SQL syntax. Bun lacks a first-class testcontainers story; the smoke
+//     script that previously covered this was deleted (one-time validation,
+//     wasn't maintained). Real fix is a Postgres service container in CI
+//     plus integration tests in the regular suite — separate piece of work.
+//   - src/migrator.ts — same reason as db.ts; SQL execution against a real DB.
+//   - src/migrations/ — SQL files, not TS.
 const EXCLUDE_PREFIXES = [
   'src/index.ts',
   'src/migrations/',
   'src/migrator.ts',
   'src/db.ts',
-  'src/api.ts',
   'tests/',
 ];
 
--- a/discord/src/commands/gankedtv.ts
+++ b/discord/src/commands/gankedtv.ts
@@ -5,6 +5,7 @@ import {
   type ChatInputCommandInteraction,
 } from 'discord.js';
 import type { Command, CommandContext } from './index.ts';
+import { ephemeral } from './replies.ts';
 import { requireManageChannels } from '../lib/permissions.ts';
 import type { Subscription } from '../db.ts';
 
@@ -65,11 +66,6 @@ const data = new SlashCommandBuilder()
   .addSubcommand((s) =>
     s.setName('resume').setDescription('Resume paused subscriptions in this channel.'),
   );
-
-const ephemeral = (content: string) => ({
-  content,
-  flags: MessageFlags.Ephemeral as MessageFlags.Ephemeral,
-});
 
 async function execute(
   interaction: ChatInputCommandInteraction,
--- a/discord/src/commands/gankedtv.ts
+++ b/discord/src/commands/gankedtv.ts
@@ -306,11 +306,19 @@ async function autocomplete(
   // than nothing. Non-empty: case-insensitive substring search via the games API.
   // Returns slug values (matching /clip autocomplete) so resolveGameId can take
   // an exact-slug shortcut over the noisier name search.
-  const games = await ctx.api.listGames({
-    search: query.length > 0 ? query : undefined,
-    limit: 25,
-    hasClips: true,
-  });
+  // Tight 2.5s timeout: Discord deadlines autocomplete at 3s — a slow games
+  // API would otherwise leave the interaction token expired (UnknownInteraction).
+  let games: { name: string; slug: string }[];
+  try {
+    games = await ctx.api.listGames({
+      search: query.length > 0 ? query : undefined,
+      limit: 25,
+      hasClips: true,
+      timeoutMs: 2500,
+    });
+  } catch {
+    games = [];
+  }
   await interaction.respond(games.slice(0, 25).map((g) => ({ name: g.name, value: g.slug })));
 }
 
--- a/discord/src/commands/gankedtv.ts
+++ b/discord/src/commands/gankedtv.ts
@@ -35,7 +35,7 @@ const data = new SlashCommandBuilder()
   .addSubcommand((s) =>
     s
       .setName('unsubscribe')
-      .setDescription('Remove a subscription from this channel.')
+      .setDescription('Remove a subscription from this channel. Pass all:true to wipe everything.')
       .addStringOption((o) =>
         o
           .setName('game')
@@ -47,6 +47,12 @@ const data = new SlashCommandBuilder()
         o
           .setName('creator')
           .setDescription('Creator filter of the subscription to remove (uuid).')
+          .setRequired(false),
+      )
+      .addBooleanOption((o) =>
+        o
+          .setName('all')
+          .setDescription('Remove EVERY subscription in this channel. Cannot combine with filters.')
           .setRequired(false),
       ),
   )
@@ -182,6 +188,29 @@ async function handleUnsubscribe(
 
   const gameRaw = interaction.options.getString('game');
   const creatorRaw = interaction.options.getString('creator');
+  const all = interaction.options.getBoolean('all') ?? false;
+
+  // Reject ambiguous input rather than silently dropping the filters — a user
+  // who passed both clearly meant one of them and shouldn't get a wipe by
+  // accident.
+  if (all && (gameRaw !== null || creatorRaw !== null)) {
+    await interaction.editReply(
+      'Cannot combine `all:true` with `game`/`creator` filters. Use one or the other.',
+    );
+    return;
+  }
+
+  if (all) {
+    const removed = await ctx.db.removeAllSubscriptionsForChannel(interaction.channelId);
+    if (removed === 0) {
+      await interaction.editReply('No subscriptions in this channel.');
+      return;
+    }
+    await interaction.editReply(
+      `Removed all ${removed} subscription${removed === 1 ? '' : 's'} in this channel.`,
+    );
+    return;
+  }
 
   const { gameId, error: gameErr } = await resolveGameId(ctx, gameRaw);
   if (gameErr) {
@@ -213,20 +242,43 @@ async function handleList(
   await interaction.deferReply({ flags: MessageFlags.Ephemeral });
   const subs = await ctx.db.listSubscriptionsForChannel(interaction.channelId);
   if (subs.length === 0) {
-    await interaction.editReply('No subscriptions in this channel.');
+    await interaction.editReply(
+      'No subscriptions in this channel. Use `/gankedtv subscribe` to add one.',
+    );
     return;
   }
-  const lines = subs.map(formatSubscription);
-  await interaction.editReply(`**Subscriptions in this channel:**\n${lines.join('\n')}`);
+
+  // Single batched lookup so we can show names/slugs instead of raw numeric
+  // ids — formatSubscription falls back to the id when a game isn't in the
+  // cache (covers catalogs > listGames's 50-item cap).
+  const games = await ctx.api.listGames({ limit: 50 });
+  const byId = new Map(games.map((g) => [g.id, g] as const));
+
+  const lines = subs.map((s, i) => formatSubscription(s, i + 1, byId));
+  const count = subs.length === 1 ? '1 subscription' : `${subs.length} subscriptions`;
+  await interaction.editReply(
+    `**${count} in this channel:**\n${lines.join('\n')}\n\n` +
+      '_Remove with `/gankedtv unsubscribe game:<slug>` (or `/gankedtv unsubscribe all:true` to wipe all)._',
+  );
 }
 
-function formatSubscription(sub: Subscription): string {
+function formatSubscription(
+  sub: Subscription,
+  index: number,
+  byId: Map<number, { name: string; slug: string }>,
+): string {
   const parts: string[] = [];
-  parts.push(sub.gameId !== null ? `game=${sub.gameId}` : 'game=any');
-  parts.push(sub.creatorId !== null ? `creator=${sub.creatorId}` : 'creator=any');
-  if (sub.pingRoleId) parts.push(`ping=<@&${sub.pingRoleId}>`);
+  if (sub.gameId !== null) {
+    const g = byId.get(sub.gameId);
+    parts.push(g ? `game **${g.name}** (\`${g.slug}\`)` : `game id \`${sub.gameId}\``);
+  }
+  if (sub.creatorId !== null) parts.push(`creator \`${sub.creatorId}\``);
+  if (sub.pingRoleId) parts.push(`ping <@&${sub.pingRoleId}>`);
   if (sub.paused) parts.push('**paused**');
-  return `• ${parts.join(' · ')}`;
+  // If no filters were set this is the firehose — say so explicitly rather
+  // than printing an empty bullet.
+  const body = parts.length ? parts.join(' · ') : '_all clips_';
+  return `${index}. ${body}`;
 }
 
 async function handlePauseToggle(
--- a/discord/src/commands/gankedtv.ts
+++ b/discord/src/commands/gankedtv.ts
@@ -1,0 +1,269 @@
+import {
+  MessageFlags,
+  SlashCommandBuilder,
+  type AutocompleteInteraction,
+  type ChatInputCommandInteraction,
+} from 'discord.js';
+import type { Command, CommandContext } from './index.ts';
+import { requireManageChannels } from '../lib/permissions.ts';
+import type { Subscription } from '../db.ts';
+
+const data = new SlashCommandBuilder()
+  .setName('gankedtv')
+  .setDescription('Manage clip subscriptions for this channel.')
+  .addSubcommand((s) =>
+    s
+      .setName('subscribe')
+      .setDescription('Subscribe this channel to new GankedTV clips.')
+      .addStringOption((o) =>
+        o
+          .setName('game')
+          .setDescription('Only post clips for this game.')
+          .setAutocomplete(true)
+          .setRequired(false),
+      )
+      .addStringOption((o) =>
+        o
+          .setName('creator')
+          .setDescription('Only post clips from this creator (uuid).')
+          .setRequired(false),
+      )
+      .addRoleOption((o) =>
+        o.setName('ping_role').setDescription('Role to ping on each post.').setRequired(false),
+      ),
+  )
+  .addSubcommand((s) =>
+    s
+      .setName('unsubscribe')
+      .setDescription('Remove a subscription from this channel.')
+      .addStringOption((o) =>
+        o
+          .setName('game')
+          .setDescription('Game filter of the subscription to remove.')
+          .setAutocomplete(true)
+          .setRequired(false),
+      )
+      .addStringOption((o) =>
+        o
+          .setName('creator')
+          .setDescription('Creator filter of the subscription to remove (uuid).')
+          .setRequired(false),
+      ),
+  )
+  .addSubcommand((s) =>
+    s.setName('subscriptions').setDescription('List subscriptions for this channel.'),
+  )
+  .addSubcommand((s) =>
+    s.setName('pause').setDescription('Pause all subscriptions in this channel.'),
+  )
+  .addSubcommand((s) =>
+    s.setName('resume').setDescription('Resume paused subscriptions in this channel.'),
+  );
+
+const ephemeral = (content: string) => ({
+  content,
+  flags: MessageFlags.Ephemeral as MessageFlags.Ephemeral,
+});
+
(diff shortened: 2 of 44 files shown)


discord/scripts/check-coverage.ts
```diff
@@ -0,0 +1,105 @@
...
+// Files included in test discovery but excluded from the coverage denominator —
+// kept in sync with bunfig.toml's coveragePathIgnorePatterns. Listed by prefix
+// match against the SF: path emitted by bun's lcov reporter.
+const EXCLUDE_PREFIXES = [
+  'src/index.ts',
+  'src/migrations/',
+  'src/migrator.ts',
+  'src/db.ts',
+  'src/api.ts',
+  'tests/',
```
mintycw, line 48: **Important — coverage gate excludes the heaviest production files.** The exclusion list (mirrored in `bunfig.toml`) drops `src/db.ts` (199 LoC), `src/api.ts` (124 LoC), `src/migrator.ts` (54 LoC), and `src/index.ts` (161 LoC) from the denominator — ~540 lines of the most failure-prone code (DB queries, HTTP client, migration runner, lifecycle wiring). The 99.81% / 98.72% headline numbers in the PR description are computed against the *complement* of this list, which is mostly pure-function code (poller, filters, commands, replies). The project's own CLAUDE.md flags this pattern for `Program.cs`: *"Keep `Program.cs` to pure DI wiring + config binding; any real logic belongs in a service so it stays inside the coverage denominator."* `api.ts` is fully pure (URL building + fetch + JSON parse) and trivially testable; the helpers in `db.ts` are pure SQL + parameter binding. Suggest moving both into the gate, and only excluding genuine bootstrap/IO wrappers.

discord/src/poller.ts
```diff
@@ -0,0 +1,219 @@
...
+      const ok = await postOne(clip, sub, db, fanout, log);
+      if (ok) posts++;
+    }
+  }
+
+  // Cursor moves to the newest clip we processed. If any post failed for a
+  // (clip, sub) pair, the failure was logged but doesn't block the cursor —
+  // the post-log dedupe handles retry next round when isPosted() returns
+  // false because we never recorded.
+  const newest = fresh[fresh.length - 1];
```
mintycw: **Important — partial fanout failures lose clips permanently.** After the fanout loop, the cursor is advanced unconditionally to `fresh[fresh.length - 1].createdAt`. If any `postOne` returned `false` (channel down, Missing Access, rate limit), that (channel, clip) pair was *not* written to `discord_post_log` and the cursor has now moved past it. Next poll filters `createdAt >= newest.createdAt`, so the failed clip is never re-fetched and never retried.

The `isPosted()` pre-check + at-least-once delivery comment two lines above only addresses the *opposite* failure mode (record-post crash after a successful send) — it does nothing for fanout failures because the API won't return those clips again.

Two concrete fixes:
1. Track per-clip success across all matched subs and only advance the cursor as far as the newest clip whose subs all succeeded-or-were-deduped.
2. Write a `discord_post_failed` retry queue keyed on (channel, clip) and have the poller drain it alongside the feed.

A flaky channel (e.g. bot temporarily missing perms after a role change) will lose every clip posted during the outage with the current design.

mintycw (commented):
## PR #111 — Discord bot review

Nice, focused service. The poller has a clear high-water-mark design with conscious tradeoffs documented in comments, the SQL is uniformly parametrized through tagged templates, and the off-by-default boot path mirrors `IgdbSyncHostedService` cleanly. That said, a few items deserve attention before merge — most importantly the cursor-advances-on-fanout-failure semantics, the coverage gate excluding the heaviest files, and an advisory-lock acquisition pattern that may not behave as intended.

### Strengths

- `discord/src/migrations/001_initial.sql` — `UNIQUE NULLS NOT DISTINCT` choice is correct and the rationale is documented. The split idxs (`WHERE game_id IS NOT NULL` / `WHERE creator_id IS NOT NULL`) keep the index size proportional to filtered subs only.
- `discord/src/db.ts:78` — `IS NOT DISTINCT FROM` for NULL-safe equality is the right tool; avoids the verbose OR-IS-NULL idiom.
- `discord/src/poller.ts:58` — guarding against garbage cursor values with re-anchor is a real production-saver; flag this as a behavior to monitor in logs.
- `discord/src/poller.ts:163` — reentrancy guard on `setInterval` is correct and necessary.
- `discord/src/index.ts:64` — interaction error handler correctly distinguishes deferred-vs-not state and uses `editReply` to avoid `InteractionAlreadyReplied`.
- `discord/src/db.ts:181` — `statement_timeout` + `idle_in_transaction_session_timeout` is a nice belt-and-braces config that's easy to forget.
- `discord/Dockerfile` runs as non-root user.
- `.github/workflows/discord.yml` correctly scopes `permissions: contents: read` and uses `persist-credentials: false`.

### Critical

None.

### Important

1. **Coverage gate excludes the heaviest production files** — `discord/scripts/check-coverage.ts:24-31` and `discord/bunfig.toml`. The 99.81% line / 98.72% function numbers in the PR description are computed against a denominator that excludes `src/db.ts` (199 LoC), `src/api.ts` (124 LoC), `src/migrator.ts` (54 LoC), and `src/index.ts` (161 LoC) — i.e. ~540 lines of the bot's most failure-prone code (DB queries, HTTP client, migration runner, lifecycle wiring). The remaining covered surface is roughly the poller + filters + commands + replies. Excluding the entry-point is reasonable (matches the server's `Program.cs` carveout) but **excluding `db.ts` and `api.ts` is the same anti-pattern called out in the project's CLAUDE.md** ("Keep `Program.cs` to pure DI wiring + config binding; any real logic belongs in a service so it stays inside the coverage denominator"). At minimum, `api.ts` (pure transform of fetch responses) and the pure helpers in `db.ts` should be inside the gate. Headline % should reflect what's actually tested.

2. **Partial-fanout failures silently drop clips for the affected channels** — `discord/src/poller.ts:127-129`. After fanout, the cursor advances to `fresh[fresh.length - 1].createdAt` unconditionally. If `postOne` returned false for a (clip, sub) pair (channel went down, Missing Access, rate limit, etc.), the clip is *not* recorded in `discord_post_log` — but the cursor has moved past it, so the next poll's `getFeed({cursor: ...})` won't include it. The `isPosted()` pre-check + at-least-once delivery only protects against the *opposite* failure (recordPost crash after a successful send). End result: a flaky channel loses clips permanently rather than retrying. Two practical fixes:
   - Only advance the cursor to the newest clip for which *all* matched subscriptions either succeeded or were skipped via `isPosted()`, OR
   - Add a `discord_post_failed` retry queue keyed by (channel, clip) that the poller consumes alongside the feed.

3. **Migration advisory lock may not be held by the same session that runs the DDL** — `discord/src/migrator.ts:32-50`. `sql\`SELECT pg_advisory_lock(...)\`` and `sql.begin(...)` and `sql\`SELECT pg_advisory_unlock(...)\`` are three separate calls against a connection *pool*. porsager/postgres makes no guarantee they land on the same backend. Effects:
   - Two starting bot instances can both acquire the lock on different backends (one pool's connection, another pool's connection) — they will *not* serialize, because PG advisory locks are session-scoped and these are two different sessions on two different backends. The `pg_advisory_lock` blocks within a single backend only.
   - `pg_advisory_unlock` only releases locks held by the *calling* session; if it lands on a different pool connection than the one that took the lock, the unlock is a silent no-op and the lock persists until that backend's session terminates.
   - Fix: `await sql.reserve()` to pin one connection for the lock+migrate+unlock window (porsager/postgres provides this), or use a transaction-scoped `pg_advisory_xact_lock` inside `sql.begin()` so the lock auto-releases at commit/rollback.

4. **Stale subscriptions for deleted/kicked channels generate errors every poll forever** — `discord/src/index.ts:138`. The `Guilds` intent is the only one declared, and there's no listener for `channelDelete` / `guildDelete`. When a channel is deleted or the bot is kicked, `client.channels.fetch(target.channelId)` throws `DiscordAPIError[10003]` (Unknown Channel) — `postOne` logs and returns false, the cursor still advances, and this repeats for every new clip indefinitely. Cheap mitigation: wire `channelDelete` and `guildDelete` to `removeAllSubscriptionsForChannel`, and treat 10003 / 10004 / 50001 specifically inside the fanout (delete the offending subscription).

5. **The `enabled` env contract differs from what the PR description claims** — `discord/src/config.ts:32`. The description says "Off by default — boots and exits cleanly without `DISCORD_BOT_TOKEN`". Actual behavior: `enabled = Boolean(token && appId)`. Token alone is insufficient; the app id is also required. The `.env.example` ships both as empty, so default behavior is correct, but the error mode for "token set, app id unset" is *silent* exit with `disabled` log message instead of a misconfig error. Suggest logging at WARN with explicit "DISCORD_BOT_TOKEN set but DISCORD_BOT_APP_ID missing" so an operator who set one but forgot the other isn't left wondering why the bot is no-op.

### Minor

1. **Fanout comment is wrong about `client.channels.fetch` semantics** — `discord/src/index.ts:135-137`. The comment says "Returns null if the bot was kicked or the channel was deleted". discord.js v14 actually **throws** `DiscordAPIError` for permission/missing errors; only forced-cache misses return null. The try/catch in `postOne` handles both paths today, but the comment misleads future maintainers debugging fanout silence.

2. **`getFeed` autocomplete can exceed Discord's 3 s deadline** — `discord/src/commands/{gankedtv,clip}.ts` autocomplete handlers call `ctx.api.listGames(...)` with a 10 s fetch timeout. If the API takes > 3 s, `interaction.respond` throws `UnknownInteraction`. Either tighten the autocomplete-specific timeout (e.g. 2 s) or `Promise.race` against a 2.5 s deadline and return `respond([])` on timeout.

3. **`pageWalkedPastCursor` uses strict `<` while the freshness filter uses `>=`** — `discord/src/poller.ts:91-95`. Asymmetric by design (boundary inclusion + post-log dedupe), but with `maxPages` only at 10 and a heavily-batched insert all at the exact same `createdAt`, the loop could traverse `maxPages` worth of identical-timestamp clips without ever satisfying `lastItem.createdAt < lastSeen`. Unlikely in practice (seed scripts aside), but a hard cap on the loop is the right safety net, which you have. Worth a code comment that the boundary-equality case is the loop-exit-by-maxPages path.

4. **`removeAllSubscriptionsForChannel` has no confirmation prompt** — `discord/src/commands/gankedtv.ts:194`. `ManageChannels` is the protection. A double-confirm modal (`buttons` with a 30 s timeout) is cheap insurance for a destructive op, but optional.

5. **`postToChannel` discards Discord rate-limit info** — `discord/src/posting.ts:31`. `channel.send` rejects with `DiscordAPIError` on rate limit; we lose the `retry_after` value. Not a blocker (discord.js auto-retries some rate limits), but logging it would help diagnose poll-cycle throughput problems.

6. **`parseEnvFile` doesn't strip `export ` prefix** — `discord/src/loadEnv.ts:18`. A user who runs `export FOO=bar` inside `.env` (a common dotenv extension) would get a key of `export FOO`. Minor; document or strip.

7. **PG 15+ requirement not documented in CLAUDE.md** — `UNIQUE NULLS NOT DISTINCT` is PG 15+. The compose file pins PG 18 so dev is fine, but operators running against shared older Postgres clusters will see migration failure with a confusing error. Add a one-liner in the Discord section of CLAUDE.md / .env.example.

8. **`removeSubscription` casts via `::int` / `::uuid` to disambiguate** — `discord/src/db.ts:80-82`. Without the casts, `IS NOT DISTINCT FROM` with a typed `null` parameter from JS would fail type inference. Good defensive coding; worth a one-line comment so a future reader doesn't "clean up" the casts.

### Verdict

**With fixes.** The architecture is sound and the SQL is clean. The cursor-advancement-on-failure (#2) and the migration advisory-lock pattern (#3) are real correctness issues, and the coverage gate (#1) materially changes how to read the headline coverage number. The stale-subscription cleanup (#4) is operational hygiene that'll matter the first time a channel gets deleted. None of these block a follow-up PR — they're all reachable with small, scoped changes.

_Posted by Claude (code-review subagent dispatched via `/superpowers:requesting-code-review`)._

discord/src/migrator.ts
```diff
@@ -0,0 +1,54 @@
...
+    )
+  `);
+
+  // pg_advisory_lock serialises concurrent boots so two instances can't race
+  // through the apply loop (which would otherwise double-run a migration's
+  // DDL before either gets to INSERT the filename). Session-scoped, so a
+  // crashed instance auto-releases. Lock is held only for the apply loop,
+  // not for the lifetime of the connection.
+  await sql`SELECT pg_advisory_lock(${MIGRATION_LOCK_ID})`;
+  try {
```
mintycw, line 34: **Important — advisory lock may not be held by the session that runs the DDL.** `sql\`SELECT pg_advisory_lock(...)\``, `sql.begin(...)`, and `sql\`SELECT pg_advisory_unlock(...)\`` are three separate top-level calls on a *connection pool*. porsager/postgres does not guarantee these land on the same backend.

Consequences:
- **Two simultaneous boots don't serialize** as intended. Each acquires the lock on its own pool's connection (two different sessions) — PG advisory locks only block within a single backend session. Both boots can race through `appliedSet` and the apply loop and double-run DDL. The DDL itself is `IF NOT EXISTS`-guarded so the first table create wins, but the `INSERT INTO discord_migrations` collides (PK on filename — DO NOTHING isn't on this insert), one boot crashes loudly.
- **`pg_advisory_unlock` may silently no-op.** If it lands on a different pool connection than the one that acquired the lock, it returns false and the lock persists until the original backend session terminates. Lock leak across the pool.

Fix options:
1. `const lockConn = await sql.reserve();` then run the entire lock → apply → unlock sequence through `lockConn` (then `lockConn.release()`).
2. Use `pg_advisory_xact_lock` *inside* `sql.begin()` per migration — auto-releases on commit/rollback, no manual unlock, no cross-connection risk.

discord/src/index.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  // One channel per call. The poller invokes this per matched (clip, sub) pair
+  // and the return value (true=delivered, false=transient failure) drives
+  // whether the post-log row is written. Returning false here means the poller
+  // will isPosted()-check this pair again next round and retry.
+  const fanout: Fanout = async (clip, target) => {
+    // fetch() pulls from the cache when warm, REST when cold. Returns null if
+    // the bot was kicked or the channel was deleted — treat as a permanent
+    // failure for this round (next round will isPosted()-check + retry).
+    const channel = await client.channels.fetch(target.channelId);
```
mintycw: **Important — no cleanup of stale subscriptions for deleted channels.** Only the `Guilds` intent is declared and there's no `channelDelete` / `guildDelete` listener. When a channel is deleted or the bot is kicked, `client.channels.fetch(target.channelId)` throws `DiscordAPIError[10003]` (Unknown Channel) — `postOne` catches, logs `'post threw'`, returns false, and this repeats *for every new clip* indefinitely. The post log fills with no rows; the error log fills with one entry per clip per poll round per dead subscription.

Cheap mitigation:
- Subscribe to `channelDelete` and `guildDelete` and call `removeAllSubscriptionsForChannel(channel.id)` / iterate channels in the guild.
- In the fanout, treat `DiscordAPIError` codes `10003`, `10004`, `50001`, `50007` as terminal and delete the offending subscription (or at least mark it `paused = true`).

The cursor-advancement issue (separate comment) compounds this: every clip a dead subscription would have received is also unrecoverably lost.

discord/src/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  await client.login(config.DISCORD_BOT_TOKEN!);
+
+  // One channel per call. The poller invokes this per matched (clip, sub) pair
+  // and the return value (true=delivered, false=transient failure) drives
+  // whether the post-log row is written. Returning false here means the poller
+  // will isPosted()-check this pair again next round and retry.
+  const fanout: Fanout = async (clip, target) => {
+    // fetch() pulls from the cache when warm, REST when cold. Returns null if
+    // the bot was kicked or the channel was deleted — treat as a permanent
+    // failure for this round (next round will isPosted()-check + retry).
```
mintycw: **Minor — comment doesn't match discord.js behavior.** "Returns null if the bot was kicked or the channel was deleted" — in discord.js v14, `Client#channels#fetch` *throws* `DiscordAPIError` for permission/missing-resource errors. The `null` return is for forced-cache-miss-only fetches. Today's try/catch in `postOne` handles both paths, but the comment misleads future maintainers debugging fanout silence.

discord/src/config.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export type Config = z.infer<typeof Schema> & { enabled: boolean };
+
+export function loadConfig(env: NodeJS.ProcessEnv = process.env): Config {
+  const parsed = Schema.parse(env);
+  // Symmetric with IgdbSyncHostedService: presence of the token is the on-switch.
+  // Without it, boot logs "disabled" and exits — the compose service still starts,
+  // it just no-ops. App ID is required alongside the token because slash command
+  // registration uses the REST API and addresses commands by application id.
+  const enabled = Boolean(parsed.DISCORD_BOT_TOKEN && parsed.DISCORD_BOT_APP_ID);
```
mintycw: **Important — silent disable on partial config.** `enabled = Boolean(token && appId)`. The PR description and `.env.example` describe the off-switch as just `DISCORD_BOT_TOKEN` — but in reality you can ship token-set + app-id-unset and the bot will silently `disabled; exiting` with no signal that you misconfigured.

Suggest:
```ts
const hasToken = Boolean(parsed.DISCORD_BOT_TOKEN);
const hasAppId = Boolean(parsed.DISCORD_BOT_APP_ID);
if (hasToken !== hasAppId) {
  console.warn(JSON.stringify({
    level: 'warn',
    msg: 'partial Discord config — bot will not start',
    hasToken, hasAppId,
  }));
}
const enabled = hasToken && hasAppId;
```
Logs an explicit signal at boot when an operator set one but forgot the other.

discord/src/commands/gankedtv.ts
```diff
@@ -0,0 +1,317 @@
...
+}
+
+async function autocomplete(
+  interaction: AutocompleteInteraction,
+  ctx: CommandContext,
+): Promise<void> {
+  const focused = interaction.options.getFocused(true);
+  if (focused.name !== 'game') return;
+  const query = focused.value?.toString() ?? '';
+  // Empty input: show first 25 games-with-clips so users see *something* rather
```
mintycw, line 305: **Minor — autocomplete can exceed Discord's 3 s deadline.** This handler calls `ctx.api.listGames(...)` with the default 10 s fetch timeout from `api.ts`. If the API responds in > 3 s, `interaction.respond` throws `UnknownInteraction` (interaction token expired).

Either pass an autocomplete-specific shorter timeout to the API client (~2 s), or `Promise.race` against a 2.5 s deadline and `respond([])` on timeout. Same applies to `discord/src/commands/clip.ts` autocomplete.

discord/src/migrations/001_initial.sql
```diff
@@ -0,0 +1,71 @@
...
+    created_by     text NOT NULL,
+    -- One rule per (channel, game, creator) combo. NULLS NOT DISTINCT (PG15+)
+    -- is the key choice: two `/subscribe` calls with no filters on the same
+    -- channel would otherwise create two phantom firehose rows (default
+    -- NULLS DISTINCT treats (C, NULL, NULL) and (C, NULL, NULL) as distinct).
+    -- The semantics we want: (C, NULL, NULL) collides with itself, but
+    -- (C, 42, NULL) is still distinct from (C, NULL, NULL) because 42 ≠ NULL.
+    UNIQUE NULLS NOT DISTINCT (channel_id, game_id, creator_id)
+);
+
```
mintycw, line 47: **Minor — PG 15+ requirement should be documented.** `UNIQUE NULLS NOT DISTINCT` requires PostgreSQL 15+. The compose stack runs PG 18 so dev is fine, but operators running the bot against a shared older Postgres cluster will hit a confusing migration failure. Add a one-liner to CLAUDE.md / .env.example explicitly stating the minimum PG version.

discord/scripts/check-coverage.ts
Turbootzz: Applied — `src/api.ts` is now in the coverage denominator. Added [discord/tests/api.test.ts](https://github.com/gankedtv/gankedtv/blob/fa3c41e/discord/tests/api.test.ts) covering URL building, query encoding, error responses, and timeout via an injected `fetchImpl` parameter to `createApi`. New gate run: 117 tests, 99.40% line / 98.86% function with `api.ts` included.

Sharpened the exclude-list comment in `scripts/check-coverage.ts` to explain why `db.ts` + `migrator.ts` are still out (real Postgres needed; testcontainers in Bun is the right long-term fix as a separate piece of work) vs. things we should never exclude.

discord/src/poller.ts
Turbootzz: Applied — cursor now halts at the first-failure boundary instead of unconditionally advancing. Implementation in [discord/src/poller.ts](https://github.com/gankedtv/gankedtv/blob/fa3c41e/discord/src/poller.ts):

- `postOne` returns `'sent' | 'skipped' | 'failed'` (was `boolean`, which conflated "already posted" and "send failed").
- `pollOnce` tracks `firstFailureIdx` across all clips and advances the cursor only to `fresh[firstFailureIdx - 1]` on partial failure, or skips the advance entirely if the very first clip failed.
- Later clips in the round are still processed so we don't artificially halt newer work; their `isPosted` pre-check short-circuits next round for the ones that already succeeded.

Picked option 1 over the `discord_post_failed` retry queue — keeps the schema lean and the retry semantics derive naturally from the existing cursor + post-log invariants. Combined with M4 (dead-channel auto-cleanup) so a permanently-dead channel doesn't poison the cursor forever.

Tests added: partial mid-round failure, first-clip failure (no advance), partial multi-sub failure on same clip.

discord/src/migrator.ts
Turbootzz: Applied — picked option 1 (`sql.reserve()`). Implementation in [discord/src/migrator.ts](https://github.com/gankedtv/gankedtv/blob/fa3c41e/discord/src/migrator.ts):

- Single backend connection pinned for the entire lock → apply → unlock sequence.
- `conn.begin(...)` for the per-migration tx inherits the same session, so the advisory lock covers the DDL too.
- Explicit `conn.release()` in `finally` so a thrown migration doesn't leak the pool slot.

Chose `pg_advisory_lock` over `pg_advisory_xact_lock` per-migration so the lock spans the whole apply loop (not just one migration's tx) — prevents a second instance from racing in between the loop's individual transactions.

discord/src/index.ts
Turbootzz: Applied — dead-channel auto-cleanup in the fanout. Implementation in [discord/src/index.ts](https://github.com/gankedtv/gankedtv/blob/fa3c41e/discord/src/index.ts):

- `client.channels.fetch()` wrapped in try/catch. On `DiscordAPIError` codes `10003`, `10004`, `50001`, `50007`: call `db.removeAllSubscriptionsForChannel(channelId)`, log a warn with the code + removed count, return `false`.
- Returning `false` triggers M2's cursor halt this round; next round the subs are gone so the cursor advances cleanly without retrying.

Lazy cleanup (discover-on-send) rather than event listeners (channelDelete / guildDelete) for now — covers the same surface with one place to maintain. Event listeners are a follow-up if log noise becomes an issue.

discord/src/index.ts
Turbootzz: Applied — rewrote the comment to match v14 behavior. [discord/src/index.ts](https://github.com/gankedtv/gankedtv/blob/fa3c41e/discord/src/index.ts) now states explicitly that `Client#channels#fetch` throws `DiscordAPIError` for missing/inaccessible resources (codes 10003, 10004, 50001, 50007 etc.) and `null` is only for forced cache-miss configurations. The try/catch a few lines down is now the documented handler for both paths, and it's wired into M4's dead-channel cleanup.

discord/src/config.ts
Turbootzz: Applied — partial config now emits a structured `warn` line before disable. [discord/src/config.ts](https://github.com/gankedtv/gankedtv/blob/fa3c41e/discord/src/config.ts) checks `hasToken !== hasAppId` and logs a single JSON line with both flags so it's grep-able alongside the other structured logs. The "both unset" path stays silent (intentional off-by-default).

Tests: explicit warn-line assertion, "no warn when both absent" guard, existing partial-config tests wrapped in `spyOn` so test output stays clean.

discord/src/commands/gankedtv.ts
Turbootzz: Applied — 2.5s timeout for autocomplete calls (Discord's 3s deadline minus ~500ms slack). Implementation:

- [discord/src/api.ts](https://github.com/gankedtv/gankedtv/blob/fa3c41e/discord/src/api.ts): added `timeoutMs` to `RequestOptions`, threaded through every method.
- Autocomplete handlers in [discord/src/commands/gankedtv.ts](https://github.com/gankedtv/gankedtv/blob/fa3c41e/discord/src/commands/gankedtv.ts) and [discord/src/commands/clip.ts](https://github.com/gankedtv/gankedtv/blob/fa3c41e/discord/src/commands/clip.ts) pass `timeoutMs: 2500` and try/catch around it — on timeout we respond with `[]` rather than throw, so the interaction still ends cleanly.

Poll-loop calls keep the default 10s — they're not deadline-bound and a tighter timeout would just create transient flakiness.

discord/src/migrations/001_initial.sql
Turbootzz: Applied — added a PG15 minimum-version block to both the migration comment ([discord/src/migrations/001_initial.sql](https://github.com/gankedtv/gankedtv/blob/fa3c41e/discord/src/migrations/001_initial.sql)) and the bot section in [CLAUDE.md](https://github.com/gankedtv/gankedtv/blob/fa3c41e/CLAUDE.md). The migration comment also describes the fallback (partial unique index + app-side guards) for operators stuck on PG ≤ 14.

mintycw (approved):
Reviewed via `/superpowers:requesting-code-review`. No Critical issues; the risky areas (at-least-once delivery, cursor halting on partial failures, advisory-lock migrations, dead-channel cleanup) are all sound and well-tested.

Approving. Important issues found are scaling/UX polish rather than correctness blockers — tracking as follow-ups rather than holding the merge:

1. `listAllSubscriptions` is unfiltered per poll round — scales linearly with sub count. Push filters down to SQL when this matters.
2. Overlapping subscriptions on the same channel: `pingRoleId` is decided by `ORDER BY created_at ASC` in `matchingSubscriptions`. Worth either collapsing matches per `(channel, clip)` preferring the most-specific filter, or documenting the behavior on `/gankedtv subscriptions`.
3. `loadEnv.ts` precedence comment overpromises — the `discord/.env` layer only loads when launched from `discord/`. Either tighten the doc or load it from a path resolved relative to the source file.
4. `discord_post_log` grows unbounded with no pruning. File a retention plan now so the future migration just flips a switch.
5. Dead-channel cleanup re-runs once per matching sub in the same poll round. Memoize a per-round set of channels already cleaned up.

Strengths worth calling out: `sql.reserve()` on the advisory lock, `UNIQUE NULLS NOT DISTINCT` for firehose dedupe, the reentrancy guard on `setInterval`, the explicit dead-channel error code list (10003/10004/50001/50007), and the tamper-resistant coverage gate script. No `Co-Authored-By: Claude` trailers across the 7 commits.

Issue #87, opened 2026-05-16
[server][web] Trending: view counting + time-weighted feed + wire TrendingView tabs


Turbootzz:
## Summary

Phase 3 (Discovery) — make trending real. Today [TrendingView.vue](web/src/views/TrendingView.vue) loads the public feed, sorts it in-memory by `likeCount`, and renders hardcoded trend arrows (lines 21–23 carry the TODO). There's also no way to record a view — `clips.view_count` exists in the schema but no endpoint increments it. This issue adds view counting, a time-weighted trending feed, and wires the Trending view's time-window tabs to real data.

Per [PLAN.md §9](PLAN.md) Phase 3: *Trending: time-weighted scoring (likes + views in last 24h)*.

### View counting

New `POST /clips/{id}/view` endpoint (likely on `ClipsReadEndpoints.cs` or a new `ClipsViewsEndpoints.cs`):

- Anonymous-friendly (no auth required).
- Dedup window: skip same `(clip_id, user_id | ip_hash)` within 30 min. Use `IMemoryCache` for v1 (Redis later, per [PLAN.md §4](PLAN.md)).
- Increments `clips.view_count` atomically (`UPDATE clips SET view_count = view_count + 1 WHERE id = @id`).
- IP-rate-limited (20/min) using the [AuthRateLimiting](server/src/GankedTV.Api/Auth/AuthRateLimiting.cs) pattern.
- Returns 204 on success and on duplicate (silently no-op).

### Trending feed

Extend `GET /clips/feed` with `sort=trending&window=24h|7d` (default still `latest`):

**Score formula (Reddit-style, tunable):**

```
score = (likes_in_window * 3 + views_in_window) / pow(hours_since_post + 2, 1.5)
```

**Implementation approach is the implementer's call** — pick whichever fits your taste; both are acceptable:

- **Option A — cached hosted-service.** New `IHostedService` runs every 5 min, queries likes/views per window from the existing `likes` table (and `clips.view_count` snapshot deltas), computes top-N per `(window)` key, caches in-process. Trending feed reads from the cache. Pros: no new tables, fits PLAN.md's "Redis later" trajectory, lowest write overhead. Cons: 5-min staleness, lost on restart.
[...]

PR #112, opened 2026-05-23
[server][infra] Redis: hot-feed/trending cache + cluster-wide rate limiting

### LINKED ISSUE #101: [server][web] Redis: hot-feed cache + cluster-wide rate limiting + trending cache migration
## Summary

Phase 4 (Scale & Grow) — introduce Redis as the shared cache + cross-instance coordination layer. Today there is **no `IMemoryCache`/`IDistributedCache` or Redis** anywhere in the API — every read hits Postgres fresh, rate limiters are in-process per-pod, and #87's trending feed deliberately caches in-process "for v1" with the intent to move to Redis.

Per [PLAN.md §9](PLAN.md) Phase 4: *Redis caching for hot feeds*. This issue also resolves the Phase-4 TODO at [ClipsRateLimiting.cs](server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs) (line ~28: *"Phase 4 moves to a Redis-backed limiter for cluster-wide enforcement"*).

**Soft dependency:** lands cleanest after #87 (trending) so its in-process cache can be migrated in the same change.

## Infra

- Add a `redis:7` service to [docker-compose.dev.yml](docker-compose.dev.yml) (healthcheck via `redis-cli ping`, named volume `redis_data`, host port default `6379`).
- Add `REDIS_URL` to [.env.example](.env.example).
- Add a worktree port offset for Redis in [scripts/new-worktree.sh](scripts/new-worktree.sh) and surface it in `make ports`.

## Server

- Add `StackExchange.Redis` + register `IConnectionMultiplexer` + `IDistributedCache` (Redis). Bind a `RedisOptions` from `REDIS_URL` in `Program.cs` (config binding only — keep logic in services).
- **Graceful degradation:** if `REDIS_URL` is unset/unreachable, fall back to in-process (`IMemoryCache`) so local dev needs no Redis — mirrors the IGDB-optional pattern (`IgdbSyncHostedService`). No-op without it; never 500.
- **Migrate the trending cache** (#87) from in-process to the shared Redis cache so multiple pods agree on the trending list.
- **Cache hot read paths** behind a short TTL (~30–60s) with write-time invalidation:
  - `GET /clips/feed` first page (`sort=latest`, no cursor) — [ClipsReadEndpoints.cs](server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs) `BuildFeedPageAsync`.
  - `GET /games/{slug}/clips` first page — [GamesEndpoints.cs](server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs).
  - Invalidate on clip create / delete / visibility change. **Do not cache** authenticated/personalized fields (`likedByMe`) — cache only the anonymous projection or vary the key by caller.
- **Migrate rate limiters** ([ClipsRateLimiting.cs](server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs), [AuthRateLimiting.cs](server/src/GankedTV.Api/Auth/AuthRateLimiting.cs)) to a Redis-backed fixed/sliding window so limits are enforced cluster-wide instead of per-pod. Keep the same partition keys (`u:{userId}` / `ip:{ip}`) and the 429 + `Retry-After` envelope.

## How to test manually

```bash
make up   # now also starts redis
# Hot feed cache
curl -s 'http://localhost:5050/clips/feed' >/dev/null         # cold: query Postgres
redis-cli -u "$REDIS_URL" keys '*'                            # feed key present
# within TTL the second call is served from Redis (verify via logs / EXPLAIN traffic)

# Cluster-wide rate limit (two API instances behind one Redis)
# fire >30 clip writes/min split across instances → 429 from both

# Graceful degradation
docker compose -f docker-compose.dev.yml stop redis
curl -s 'http://localhost:5050/clips/feed' | jq '.items | length'   # still 200, falls back in-process
```

## Checklist

- [ ] Feed first page served from Redis within TTL (no Postgres hit); cursor pages bypass cache
- [ ] Cache invalidated on new clip / delete / visibility change (stale entry not served)
- [ ] Personalized fields (`likedByMe`) never leak across users via the cache
- [ ] Rate limit shared across two API instances pointed at one Redis
- [ ] Redis unavailable → falls back to in-process, no 500s
- [ ] Trending cache (#87) now reads/writes Redis

Milestone: Phase 4 — Scale & Grow · Source: [PLAN.md §9](PLAN.md) · Depends on: #87

Turbootzz:
## Summary

Introduces Redis as Phase 4's shared cache + cross-instance coordination layer. The API previously had no caching and per-pod rate limiters; this adds an optional **HybridCache** (in-memory L1 + Redis L2) over the hot read paths and a **cluster-wide Redis fixed-window rate limiter**. Everything degrades gracefully to in-process when Redis is absent, so local dev needs no Redis and a Redis outage never 500s.

## What's here

- **Infra:** `redis:7` service in `docker-compose.dev.yml`, `REDIS_HOST_PORT` default + `make ports` line, per-issue worktree offset in `new-worktree.sh`, and `REDIS_URL` in `.env.example`. Main checkout resolves to `6379`; worktrees get their offset port (verified `docker compose config` both ways).
- **Hot-feed cache** behind `IFeedCache` (HybridCache + optional Redis L2): caches the anonymous first page of the latest / per-game / per-tag feeds and trending, with a short TTL and tag-based write-time invalidation. Cursor pages and `source=following` bypass.
- **Personalization safety:** only the anonymous projection is cached; `likedByMe` is re-stamped per caller after the cache read, so a like never leaks across users.
- **Games catalog cache** (`IGamesCache`): browse list + per-game detail, TTL-only; `search` queries bypass (high-cardinality).
- **Cluster-wide rate limiting:** a shared Redis fixed-window limiter for the `clips-write` and `auth-credentials` policies — same partition keys (`u:{id}` / `ip:{ip}`) and 429 + `Retry-After` envelope — falling back to the in-process limiter on any Redis error (degraded, never fail-open).
- **Trending:** #87 shipped trending *uncached*, so this adds Redis caching to it via the same abstraction (rather than migrating an in-process cache that didn't exist).

Closes #101

## How to test manually

```bash
make up                                   # now also starts redis (healthcheck: redis-cli ping)
make ports                                # shows the redis port
# Hot-feed cache
curl -s 'http://localhost:5050/clips/feed' >/dev/null
redis-cli -u "$REDIS_URL" keys 'feed:*'   # feed:latest:20 present (also game/tag/trending after hitting those)
redis-cli -u "$REDIS_URL" keys 'games:*'  # games:list:* / games:detail:* after hitting /games
# Cluster-wide rate limit
#   fire >30 authed clip-writes/min  -> 429 + Retry-After; rl:clips-write:u:{id} key appears
# Graceful degradation
docker compose -f docker-compose.dev.yml stop redis
curl -s 'http://localhost:5050/clips/feed' | jq '.items | length'   # still 200, in-process fallback
```

## Checklist

- [x] Feed / game / tag / trending first page served from Redis within TTL; cursor pages bypass
- [x] Cache invalidated on clip ready / delete / visibility change
- [x] `likedByMe` never leaks across users via the shared cache
- [x] Rate limit shared across instances via Redis; falls back in-process when Redis is down
- [x] Redis unavailable -> in-process fallback, no 500s (smoke-tested with redis stopped)
- [x] Coverage gate passes (94% line / 86% branch)
- [x] Rebase onto latest `main` before merge (currently 1 behind — Notifications #110)

## Release Notes

* **New Features**
  * Added optional Redis support for distributed caching and cluster-wide rate limiting. Feeds and game catalogs are now cached for faster load times.
  
* **Configuration**
  * Added optional `REDIS_URL` environment variable. When configured, enables Redis-backed caching and rate limiting across instances. System gracefully falls back to in-process caching when Redis is unavailable.

* **Tests**
  * Added comprehensive tests for caching and rate limiting functionality.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/112?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

---

## Known limitations (scale-up)

- **Cross-pod cache stampede:** `HybridCache.GetOrCreateAsync` single-flights *within* a pod only. After a `RemoveByTagAsync` clears L2, N pods can each run the feed factory once on the next request (a brief thundering herd, not a correctness issue). Acceptable for v1; revisit with a backplane/lock if feed-factory cost becomes material at scale.

### DIFF
--- a/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
+++ b/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
@@ -9,6 +9,7 @@ using GankedTV.Api.Data;
 using GankedTV.Api.Data.Entities;
 using GankedTV.Api.Pagination;
 using GankedTV.Api.Problems;
+using GankedTV.Api.Services.Caching;
 using GankedTV.Api.Services.Media;
 using GankedTV.Api.Services.ObjectStorage;
 using Microsoft.EntityFrameworkCore;
@@ -101,6 +102,7 @@ public static class ClipsReadEndpoints
         GankedTvDbContext db,
         IObjectStorageService storage,
         IOptions<S3Options> s3,
+        IFeedCache feedCache,
         CancellationToken ct)
     {
         var baseQuery = db.Clips.AsNoTracking()
@@ -109,7 +111,8 @@ public static class ClipsReadEndpoints
         // `source` is treated leniently: only the literal "following" switches behaviour;
         // anything else (null, "public", garbage) falls through to the global feed. Matches
         // the same forgive-and-fall-back spirit as the cursor decoder.
-        if (string.Equals(source, "following", StringComparison.OrdinalIgnoreCase))
+        var isFollowing = string.Equals(source, "following", StringComparison.OrdinalIgnoreCase);
+        if (isFollowing)
         {
             if (!TryGetUserId(principal, out var me))
             {
@@ -140,8 +143,36 @@ public static class ClipsReadEndpoints
                 return ProblemResults.BadRequest("invalid_window");
             }
 
-            var response = await BuildTrendingFeedAsync(baseQuery, since, limit, principal, db, storage, s3, ct);
-            return Results.Ok(response);
+            // The personalised "following + trending" combination filters baseQuery per user,
+            // so it must never hit the shared cache; only the global trending feed is cached.
+            if (isFollowing)
+            {
+                var personalised = await BuildTrendingFeedAsync(baseQuery, since, limit, principal, db, storage, s3, ct);
+                return Results.Ok(personalised);
+            }
+
+            var trendingLimit = Math.Clamp(limit ?? FeedDefaultLimit, 1, TrendingMaxLimit);
+            var cachedTrending = await feedCache.GetOrCreateTrendingAsync(
+                $"feed:trending:{window}:{trendingLimit}",
+                c => new ValueTask<CachedFeedPage>(
+                    BuildAnonymousTrendingFeedAsync(baseQuery, since, limit, db, storage, s3, c)),
+                ct);
+            var trendingItems = await ApplyLikedByMeAsync(cachedTrending.Items, principal, db, ct);
+            return Results.Ok(new ClipFeedResponse(trendingItems, cachedTrending.NextCursor));
+        }
+
+        // Cache only the global latest first page (no cursor, not personalised). Cursor pages
+        // and following feeds bypass the cache and query Postgres directly.
+        if (cursor is null && !isFollowing)
+        {
+            var feedLimit = Math.Clamp(limit ?? FeedDefaultLimit, 1, FeedMaxLimit);
+            var cached = await feedCache.GetOrCreateFeedAsync(
+                $"feed:latest:{feedLimit}",
+                c => new ValueTask<CachedFeedPage>(
+                    BuildAnonymousFeedPageAsync(baseQuery, null, limit, storage, s3, c)),
+                ct);
+            var items = await ApplyLikedByMeAsync(cached.Items, principal, db, ct);
+            return Results.Ok(new ClipFeedResponse(items, cached.NextCursor));
         }
 
         var latest = await BuildFeedPageAsync(baseQuery, cursor, limit, principal, db, storage, s3, ct);
@@ -186,6 +217,24 @@ public static class ClipsReadEndpoints
         IOptions<S3Options> s3,
         CancellationToken ct)
     {
+        var page = await BuildAnonymousTrendingFeedAsync(baseQuery, since, limit, db, storage, s3, ct);
+        var items = await ApplyLikedByMeAsync(page.Items, principal, db, ct);
+        return new ClipFeedResponse(items, page.NextCursor);
+    }
+
+    // Caller-independent half of BuildTrendingFeedAsync — the scoring + rehydration + anonymous
+    // projection that FeedCache stores. The window-relative ranking is inherently approximate and
+    // self-healing, so caching it behind a short TTL (rather than invalidating on every like/view)
+    // is sufficient: a cache entry just freezes `since`/scores for one TTL.
+    internal static async Task<CachedFeedPage> BuildAnonymousTrendingFeedAsync(
+        IQueryable<Clip> baseQuery,
+        DateTimeOffset since,
+        int? limit,
+        GankedTvDbContext db,
+        IObjectStorageService storage,
+        IOptions<S3Options> s3,
+        CancellationToken ct)
+    {
         var clampedLimit = Math.Clamp(limit ?? FeedDefaultLimit, 1, TrendingMaxLimit);
 
         var candidates = await baseQuery
@@ -215,7 +264,7 @@ public static class ClipsReadEndpoints
 
         if (topIds.Count == 0)
         {
-            return new ClipFeedResponse([], NextCursor: null);
+            return new CachedFeedPage([], NextCursor: null);
         }
 
         // Re-hydrate with feed Includes (the candidate Select dropped them) and preserve
@@ -243,8 +292,8 @@ public static class ClipsReadEndpoints
             }
         }
 
-        var items = await ProjectFeedItemsAsync(ranked, principal, db, storage, s3, ct);
-        return new ClipFeedResponse(items, NextCursor: null);
+        var items = ProjectAnonymousFeedItems(ranked, storage, s3);
+        return new CachedFeedPage(items, NextCursor: null);
     }
 
     // Shared cursor-paginated feed builder. Callers pass a pre-filtered IQueryable<Clip>
@@ -256,6 +305,23 @@ public static class ClipsReadEndpoints
         int? limit,
         ClaimsPrincipal principal,
         GankedTvDbContext db,
+        IObjectStorageService storage,
+        IOptions<S3Options> s3,
+        CancellationToken ct)
+    {
+        var page = await BuildAnonymousFeedPageAsync(baseQuery, cursor, limit, storage, s3, ct);
+        var items = await ApplyLikedByMeAsync(page.Items, principal, db, ct);
+        return new ClipFeedResponse(items, page.NextCursor);
+    }
+
+    // Caller-independent half of BuildFeedPageAsync: runs the keyset query + ordering + includes
+    // + thumbnail signing into an anonymous page (no likedByMe). This is what FeedCache stores,
+    // so the cached entry never holds personalised data. No ClaimsPrincipal/DbContext needed
+    // for the likes lookup — that happens after the cache, per caller, in ApplyLikedByMeAsync.
+    internal static async Task<CachedFeedPage> BuildAnonymousFeedPageAsync(
+        IQueryable<Clip> baseQuery,
+        string? cursor,
+        int? limit,
         IObjectStorageService storage,
         IOptions<S3Options> s3,
         CancellationToken ct)
@@ -288,10 +354,10 @@ public static class ClipsReadEndpoints
         var hasMore = rows.Count > clampedLimit;
         var page = hasMore ? rows.GetRange(0, clampedLimit) : rows;
 
-        var items = await ProjectFeedItemsAsync(page, principal, db, storage, s3, ct);
+        var items = ProjectAnonymousFeedItems(page, storage, s3);
         var nextCursor = hasMore ? KeysetCursor.Build(page[^1].CreatedAt, page[^1].Id) : null;
 
-        return new ClipFeedResponse(items, nextCursor);
+        return new CachedFeedPage(items, nextCursor);
     }
 
     // Shared DTO projector for any pre-ordered, pre-included Clip list. Splits out the
@@ -305,16 +371,50 @@ public static class ClipsReadEndpoints
         IOptions<S3Options> s3,
         CancellationToken ct)
     {
+        var anonymous = ProjectAnonymousFeedItems(clips, storage, s3);
+        return await ApplyLikedByMeAsync(anonymous, principal, db, ct);
+    }
+
+    // Caller-independent projection: thumbnail signing only, LikedByMe left false. This is the
+    // shape the feed cache stores — never personalised — so one user's likes can't leak to
+    // another via a shared cache entry. likedByMe is re-stamped per request by ApplyLikedByMeAsync.
+    internal static List<ClipFeedItem> ProjectAnonymousFeedItems(
+        IReadOnlyList<Clip> clips,
+        IObjectStorageService storage,
+        IOptions<S3Options> s3)
+    {
         if (clips.Count == 0)
         {
             return [];
         }
 
-        var likedIds = await LoadLikedClipIdsAsync(db, principal, clips.Select(c => c.Id), ct);
         var thumbnailsBucket = s3.Value.ThumbnailsBucket;
         return [.. clips.Select(c => c.ToFeedItem(
             BuildThumbnailUrl(storage, thumbnailsBucket, c.ThumbnailKey),
-            likedIds.Contains(c.Id)))];
+            likedByMe: false))];
+    }
+
+    // Re-stamps the per-caller LikedByMe flag onto an anonymous (possibly cached) item list.
+    // Anonymous callers and callers with no likes among these clips get the list back unchanged
+    // (items already carry LikedByMe=false), so the only cost is one indexed Likes lookup.
+    internal static async Task<List<ClipFeedItem>> ApplyLikedByMeAsync(
+        IReadOnlyList<ClipFeedItem> items,
+        ClaimsPrincipal principal,
+        GankedTvDbContext db,
+        CancellationToken ct)
+    {
+        if (items.Count == 0)
+        {
+            return [.. items];
+        }
+
+        var likedIds = await LoadLikedClipIdsAsync(db, principal, items.Select(i => i.Id), ct);
+        if (likedIds.Count == 0)
+        {
+            return [.. items];
+        }
+
+        return [.. items.Select(i => likedIds.Contains(i.Id) ? i with { LikedByMe = true } : i)];
     }
 
     // Public Ready clips always have a thumbnail (the worker is the only path to Ready
--- a/server/src/GankedTV.Api/Services/Caching/FeedCache.cs
+++ b/server/src/GankedTV.Api/Services/Caching/FeedCache.cs
@@ -1,0 +1,81 @@
+using GankedTV.Api.Contracts.Clips;
+using Microsoft.Extensions.Caching.Hybrid;
+
+namespace GankedTV.Api.Services.Caching;
+
+/// <summary>
+/// Anonymous projection of a feed page held in the cache: the per-caller <c>LikedByMe</c>
+/// flag is deliberately left <c>false</c> here and re-stamped per request after a cache hit,
+/// so one user's likes never leak to another via the shared cache.
+/// </summary>
+public sealed record CachedFeedPage(IReadOnlyList<ClipFeedItem> Items, string? NextCursor);
+
+/// <summary>
+/// Hot read-path cache for feeds + trending. The interface lets write paths depend on (and
+/// tests verify) cache behaviour without binding to HybridCache directly.
+/// </summary>
+public interface IFeedCache
+{
+    /// <summary>Cache a feed page (latest or per-game) under the shared feed tag.</summary>
+    ValueTask<CachedFeedPage> GetOrCreateFeedAsync(
+        string key, Func<CancellationToken, ValueTask<CachedFeedPage>> factory, CancellationToken ct);
+
+    /// <summary>Cache a trending page (TTL-governed; not invalidated on writes).</summary>
+    ValueTask<CachedFeedPage> GetOrCreateTrendingAsync(
+        string key, Func<CancellationToken, ValueTask<CachedFeedPage>> factory, CancellationToken ct);
+
+    /// <summary>Drop every cached feed page (called best-effort on clip mutations).</summary>
+    ValueTask InvalidateFeedsAsync(CancellationToken ct);
+}
+
+/// <summary>
+/// Thin wrapper over <see cref="HybridCache"/> for the hot read paths (latest feed, per-game
+/// feed, trending). HybridCache is L1 in-memory + optional L2 Redis (wired in
+/// <see cref="RedisRegistration"/>): with Redis it's shared cluster-wide; without it the API
+/// still works off L1 alone. Centralises the short TTL + tag conventions so feed and trending
+/// stay DRY and invalidation has a single source of truth.
+/// </summary>
+public sealed class FeedCache(HybridCache cache) : IFeedCache
+{
+    /// <summary>Tag on every feed page (latest + per-game). One invalidation clears them all.</summary>
+    public const string FeedTag = "clips-feed";
+
+    /// <summary>Tag on trending pages. Self-heals via TTL (ranking is time-windowed/approximate),
+    /// so it's not invalidated on writes — only the short expiration governs freshness.</summary>
+    public const string TrendingTag = "trending";
+
+    // Short TTL with write-time invalidation. L1 (LocalCacheExpiration) is intentionally
+    // shorter than L2 (Expiration): with no cross-pod L1 backplane, RemoveByTagAsync clears
+    // L2 immediately but a sibling pod's L1 can serve stale until its local copy lapses —
+    // bounded to LocalCacheExpiration. Acceptable for a 30-60s hot-feed cache.
+    private static readonly HybridCacheEntryOptions Entry = new()
+    {
+        Expiration = TimeSpan.FromSeconds(60),
+        LocalCacheExpiration = TimeSpan.FromSeconds(30),
+    };
+
+    private static readonly string[] FeedTags = [FeedTag];
+    private static readonly string[] TrendingTags = [TrendingTag];
+
+    /// <summary>Cache a feed page (latest or per-game) under the <see cref="FeedTag"/>.</summary>
+    public ValueTask<CachedFeedPage> GetOrCreateFeedAsync(
+        string key,
+        Func<CancellationToken, ValueTask<CachedFeedPage>> factory,
+        CancellationToken ct) =>
+        cache.GetOrCreateAsync(key, factory, Entry, FeedTags, ct);
+
+    /// <summary>Cache a trending page under the <see cref="TrendingTag"/> (TTL-governed).</summary>
+    public ValueTask<CachedFeedPage> GetOrCreateTrendingAsync(
+        string key,
+        Func<CancellationToken, ValueTask<CachedFeedPage>> factory,
+        CancellationToken ct) =>
+        cache.GetOrCreateAsync(key, factory, Entry, TrendingTags, ct);
+
+    /// <summary>
+    /// Drop every cached feed page. Called best-effort on clip create/complete/delete and on
+    /// visibility/game changes — any of which can alter the global latest page, so we clear the
+    /// whole <see cref="FeedTag"/> rather than reason about which specific pages changed.
+    /// </summary>
+    public ValueTask InvalidateFeedsAsync(CancellationToken ct) =>
+        cache.RemoveByTagAsync(FeedTag, ct);
+}
--- a/server/src/GankedTV.Api/Services/Caching/FeedCache.cs
+++ b/server/src/GankedTV.Api/Services/Caching/FeedCache.cs
@@ -75,6 +75,12 @@ public sealed class FeedCache(HybridCache cache) : IFeedCache
     /// Drop every cached feed page. Called best-effort on clip create/complete/delete and on
     /// visibility/game changes — any of which can alter the global latest page, so we clear the
     /// whole <see cref="FeedTag"/> rather than reason about which specific pages changed.
+    /// <para>
+    /// Deliberately NOT called on like/unlike: a cached card's <c>LikeCount</c> is therefore
+    /// TTL-bounded-stale (≤ the entry's expiration), the same accepted tradeoff as trending.
+    /// Likes are high-frequency, so invalidating on each would defeat the cache; the clip detail
+    /// view and like button reflect the true count immediately.
+    /// </para>
     /// </summary>
     public ValueTask InvalidateFeedsAsync(CancellationToken ct) =>
         cache.RemoveByTagAsync(FeedTag, ct);
--- a/server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
+++ b/server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
@@ -78,6 +78,7 @@ internal sealed class RedisFixedWindowRateLimiter : RateLimiter
     private readonly Lazy<FixedWindowRateLimiter> _fallback;
     private readonly ILogger<RedisFixedWindowRateLimiter> _logger;
     private int _degradedLogged;
+    private long _lastActivityTicks = Environment.TickCount64;
 
     public RedisFixedWindowRateLimiter(
         IConnectionMultiplexer multiplexer,
@@ -95,12 +96,17 @@ internal sealed class RedisFixedWindowRateLimiter : RateLimiter
         _logger = logger;
     }
 
-    public override TimeSpan? IdleDuration => null;
+    // Report idle time since the last acquire so the partitioned limiter's reaper can evict
+    // stale per-user / per-IP limiters — returning null (the default) would pin every partition
+    // key ever seen in memory for the life of the process.
+    public override TimeSpan? IdleDuration =>
+        TimeSpan.FromMilliseconds(Environment.TickCount64 - Interlocked.Read(ref _lastActivityTicks));
 
     public override RateLimiterStatistics? GetStatistics() => null;
 
     protected override async ValueTask<RateLimitLease> AcquireAsyncCore(int permitCount, CancellationToken cancellationToken)
     {
+        Interlocked.Exchange(ref _lastActivityTicks, Environment.TickCount64);
         try
         {
             var db = _multiplexer.GetDatabase();
@@ -119,6 +125,7 @@ internal sealed class RedisFixedWindowRateLimiter : RateLimiter
 
     protected override RateLimitLease AttemptAcquireCore(int permitCount)
     {
+        Interlocked.Exchange(ref _lastActivityTicks, Environment.TickCount64);
         try
         {
             var db = _multiplexer.GetDatabase();
--- a/server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
+++ b/server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
@@ -1,0 +1,206 @@
+using System.Threading.RateLimiting;
+using Microsoft.Extensions.DependencyInjection;
+using StackExchange.Redis;
+
+namespace GankedTV.Api.Services.Caching;
+
+/// <summary>
+/// Builds the per-partition rate limiter used by the clip-write and credentials policies.
+/// When a shared <see cref="IConnectionMultiplexer"/> is registered (i.e. <c>REDIS_URL</c> is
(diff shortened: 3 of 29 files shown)


Turbootzz:
@coderabbitai review

Turbootzz:
@coderabbitai review

server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
```diff
@@ -0,0 +1,206 @@
...
+            var result = await db.ScriptEvaluateAsync(
+                Script,
+                [_key],
+                [permitCount, _windowMs]).ConfigureAwait(false);
+            return Evaluate(result);
+        }
+        catch (Exception ex) when (ex is RedisException or RedisTimeoutException)
+        {
+            LogDegraded(ex);
+            return await _fallback.Value.AcquireAsync(permitCount, cancellationToken).ConfigureAwait(false);
```
mintycw: **Important** — `ScriptEvaluateAsync` doesn't take a `CancellationToken`, so a long-running script call (or a stuck Redis) cannot be cancelled when the HTTP request is aborted — it'll pin the request until `ConfigurationOptions.AsyncTimeout` elapses. The fallback path on line 122 correctly forwards `cancellationToken`, but the happy path doesn't. Wrap the await with `WaitAsync(cancellationToken)` (`await db.ScriptEvaluateAsync(...).WaitAsync(cancellationToken)`), or set a low explicit `AsyncTimeout` in `RedisOptions.TryBuildConfiguration`, so client cancellations during a degraded Redis actually release server resources.

server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+
```
mintycw, line 132: **Important** — The catch only covers `RedisException` / `RedisTimeoutException`. If the multiplexer is disposed during shutdown (common in test teardown / graceful shutdown races), `IConnectionMultiplexer.GetDatabase()` throws `ObjectDisposedException`, which escapes here and surfaces as a 500 — defeating the "never 500 on Redis failure" guarantee. Recommend broadening to `RedisException or RedisTimeoutException or ObjectDisposedException or TimeoutException` (and mirror on line 135 in `AttemptAcquireCore`). The conventional shape is `catch (Exception ex) when (ex is not OperationCanceledException)` — anything infra-related must degrade.

server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
```diff
@@ -0,0 +1,206 @@
...
+    // limiter and the conventional Redis fixed-window tradeoff (no decrement race on rejection).
+    private const string Script = """
+        local current = redis.call('INCRBY', KEYS[1], ARGV[1])
+        local ttl = redis.call('PTTL', KEYS[1])
+        if ttl < 0 then
+          redis.call('PEXPIRE', KEYS[1], ARGV[2])
+          ttl = tonumber(ARGV[2])
+        end
+        return {current, ttl}
+        """;
```
mintycw, line 77: **Important (or upgrade the docstring to a loud warning)** — The script does `INCRBY KEYS[1] ARGV[1]` *before* the limit check. If a caller passes `permitCount=5` against a limit of 4 with current count 0, the counter jumps to 5 and every subsequent single-permit request in the same window is rejected — including the very first one. Current callers always use `permitCount=1` so this is latent, but `Create(...)` and `AcquireAsync(permitCount)` invite future misuse. Either (a) loudly document in the XML doc on `Create` that `permitCount` must be ≤ `permitLimit` or it poisons the bucket, or (b) precheck in Lua: `if redis.call('GET', KEYS[1]) + ARGV[1] > ARGV[3] then return {limit+1, ttl} end` (pass limit as a third ARGV) so over-allocations are rejected without consuming the window.

server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
```diff
@@ -0,0 +1,206 @@
...
+    private void LogDegraded(Exception ex)
+    {
+        if (Interlocked.Exchange(ref _degradedLogged, 1) == 0)
+        {
+            _logger.LogWarning(ex,
+                "Redis rate limiter unavailable for {Key}; degrading to in-process limiting until it recovers.",
+                (string?)_key);
+        }
+    }
+
```
mintycw, line 186: **Minor** — `Interlocked.Exchange(ref _degradedLogged, 1)` makes the warning fire exactly once per limiter *instance*. Because the framework caches limiter instances per partition for the process lifetime, if Redis flaps (down→up→down→up→down), you get exactly one log line for the very first outage and then silence forever. Consider resetting `_degradedLogged = 0` on the first successful `Evaluate` (just before `return RedisRateLimitLease.Acquired` on line 149), so each *new* outage logs once but recovery is observable. Also worth surfacing as a counter/metric — `ILogger` alone makes this invisible on a dashboard.

server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
```diff
@@ -0,0 +1,206 @@
...
+/// to the pre-Redis behaviour. Resolved once per partition by the policy callbacks in
+/// ClipsRateLimiting / AuthRateLimiting (the framework caches limiter instances per key).
+/// </summary>
+public sealed class RedisRateLimiterFactory(IServiceProvider services, ILoggerFactory loggerFactory)
+{
+    /// <param name="policy">Namespaces the Redis key so the two policies never share a bucket.</param>
+    /// <param name="partitionKey">The caller bucket (e.g. <c>u:{userId}</c> / <c>ip:{ip}</c>).</param>
+    public RateLimiter Create(string policy, string partitionKey, int permitLimit, TimeSpan window)
+    {
+        var multiplexer = services.GetService<IConnectionMultiplexer>();
```
mintycw, line 26: **Minor** — This runs once per partition (the framework caches limiter instances), so the perf cost is negligible — but the dependency could be a constructor parameter (`IConnectionMultiplexer? multiplexer`) since `AddGankedCaching` either registers it or doesn't. That resolves the Redis-present-vs-absent choice at DI time instead of per-call, removes the service-locator pattern, and lets `RedisRateLimiterFactoryTests` just pass `null` instead of building a `ServiceProvider`.

server/src/GankedTV.Api/Services/Caching/RedisRegistration.cs
```diff
@@ -0,0 +1,40 @@
...
+        services.AddSingleton<IGamesCache, GamesCache>();
+        services.AddSingleton<RedisRateLimiterFactory>();
+
+        if (options.TryBuildConfiguration(out var configuration))
+        {
+            // AbortOnConnectFail=false (set in TryBuildConfiguration) means Connect returns a
+            // multiplexer immediately even if Redis is down at boot; it reconnects in the
+            // background. A single shared multiplexer backs both the L2 cache and the limiter.
+            var multiplexer = ConnectionMultiplexer.Connect(configuration);
+            services.AddSingleton<IConnectionMultiplexer>(multiplexer);
```
mintycw: **Important** — `services.AddSingleton<IConnectionMultiplexer>(multiplexer)` (instance overload) does **not** register the instance for container disposal — MS.DI only disposes services the container *created*. The multiplexer will leak on `WebApplication.StopAsync` (real consequences: Redis sees abrupt TCP drops instead of a clean QUIT on pod shutdown; integration tests that spin multiple `WebApplicationFactory` instances accumulate background reconnect threads). Switch to a factory registration so the container owns it:

```csharp
services.AddSingleton<IConnectionMultiplexer>(_ => ConnectionMultiplexer.Connect(configuration));
services.AddStackExchangeRedisCache(o =>
    o.ConnectionMultiplexerFactory = () => Task.FromResult(
        services.BuildServiceProvider().GetRequiredService<IConnectionMultiplexer>()));
```

(or capture the configuration in a closure and use the same factory for both registrations to keep a single instance).

server/src/GankedTV.Api/Services/Caching/RedisRegistration.cs
```diff
@@ -0,0 +1,40 @@
...
+        services.AddSingleton<IFeedCache, FeedCache>();
+        services.AddSingleton<IGamesCache, GamesCache>();
+        services.AddSingleton<RedisRateLimiterFactory>();
+
+        if (options.TryBuildConfiguration(out var configuration))
+        {
+            // AbortOnConnectFail=false (set in TryBuildConfiguration) means Connect returns a
+            // multiplexer immediately even if Redis is down at boot; it reconnects in the
+            // background. A single shared multiplexer backs both the L2 cache and the limiter.
+            var multiplexer = ConnectionMultiplexer.Connect(configuration);
```
mintycw: **Important** — Even with `AbortOnConnectFail=false`, `ConnectionMultiplexer.Connect` performs a synchronous handshake attempt and only returns once the multiplexer has tried at least once. With a slow / DNS-unresponsive Redis this stretches boot by `ConnectTimeout` (default 5s) per pod — a worst-case 5s startup penalty for every pod whenever Redis is degraded at boot. Cheapest fix: set `configuration.ConnectTimeout = 1000;` in `RedisOptions.TryBuildConfiguration`. Stronger fix: do `ConnectAsync` from a background `IHostedService` so the multiplexer is lazily attached and startup never blocks on Redis.

server/src/GankedTV.Api/Services/Caching/FeedCache.cs
```diff
@@ -0,0 +1,81 @@
...
+        CancellationToken ct) =>
+        cache.GetOrCreateAsync(key, factory, Entry, TrendingTags, ct);
+
+    /// <summary>
+    /// Drop every cached feed page. Called best-effort on clip create/complete/delete and on
+    /// visibility/game changes — any of which can alter the global latest page, so we clear the
+    /// whole <see cref="FeedTag"/> rather than reason about which specific pages changed.
+    /// </summary>
+    public ValueTask InvalidateFeedsAsync(CancellationToken ct) =>
+        cache.RemoveByTagAsync(FeedTag, ct);
```
mintycw, line 86: **Important — invalidation surface gap.** `ClipFeedItem` carries `LikeCount`, which is part of the *anonymous* cached projection. Like/unlike updates `Clip.LikeCount` in Postgres but does **not** call `InvalidateFeedsAsync`, so cached feed pages show stale counts for up to 60s (L2) / 30s (L1). Trending explicitly documents "self-heals via TTL" (line 43-45) — the latest / per-game / per-tag feeds either need an analogous docstring acceptance of count lag, or `LikesEndpoints.LikeClip` / `UnlikeClip` need to fire `feedCache.InvalidateFeedsAsync` on success. Right now the doc reads as if write-time invalidation keeps counts fresh, but it doesn't for likes.

server/src/GankedTV.Api/Services/Caching/FeedCache.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    // Short TTL with write-time invalidation. L1 (LocalCacheExpiration) is intentionally
+    // shorter than L2 (Expiration): with no cross-pod L1 backplane, RemoveByTagAsync clears
+    // L2 immediately but a sibling pod's L1 can serve stale until its local copy lapses —
+    // bounded to LocalCacheExpiration. Acceptable for a 30-60s hot-feed cache.
+    private static readonly HybridCacheEntryOptions Entry = new()
+    {
+        Expiration = TimeSpan.FromSeconds(60),
+        LocalCacheExpiration = TimeSpan.FromSeconds(30),
+    };
```
mintycw, line 55: **Minor** — No per-key size cap. A pathological feed page (large titles/descriptions × max items × many tags) gets stored in L1 without a size guard; with HybridCache backing on `IMemoryCache`, unbounded payload growth silently consumes server RAM. Not load-bearing at current scale, but worth setting `MaximumPayloadBytes` on `HybridCacheOptions` (when registering `AddHybridCache`) to fail fast on outlier entries. Default is 1MiB which is probably fine — just make it explicit so it's a deliberate choice.

server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
```diff
@@ -88,8 +91,36 @@ private static async Task<IResult> GetFeed(
...
+            {
+                var personalised = await BuildTrendingFeedAsync(baseQuery, since, limit, principal, db, storage, s3, ct);
+                return Results.Ok(personalised);
+            }
+
+            var trendingLimit = Math.Clamp(limit ?? FeedDefaultLimit, 1, TrendingMaxLimit);
+            var cachedTrending = await feedCache.GetOrCreateTrendingAsync(
+                $"feed:trending:{window}:{trendingLimit}",
+                c => new ValueTask<CachedFeedPage>(
+                    BuildAnonymousTrendingFeedAsync(baseQuery, since, limit, db, storage, s3, c)),
```
mintycw, line 158: **Minor** — The cache key uses `trendingLimit` (clamped), but the factory receives raw `limit` (nullable, unclamped). The inner builder clamps again with the same formula so behaviour is identical — just surprising to read because key and payload appear to use different inputs. Pass `trendingLimit` into the factory (`BuildAnonymousTrendingFeedAsync(baseQuery, since, trendingLimit, ...)`) so the cache-key/payload alignment is obvious at a glance. Same pattern at line 120 for the latest feed (`feedLimit` vs `limit`).

server/tests/GankedTV.Api.Tests/Services/Caching/RedisRateLimiterFactoryTests.cs
```diff
@@ -0,0 +1,114 @@
...
+    // {current, ttlMs} multi-bulk reply, matching the Lua script's return shape.
+    private static RedisResult ScriptReply(long current, long ttlMs) =>
+        RedisResult.Create(new RedisValue[] { current, ttlMs });
+
+    private static (IConnectionMultiplexer mux, IDatabase db) MuxReturning(RedisResult reply)
+    {
+        var db = Substitute.For<IDatabase>();
+        db.ScriptEvaluateAsync(Arg.Any<string>(), Arg.Any<RedisKey[]>(), Arg.Any<RedisValue[]>(), Arg.Any<CommandFlags>())
+            .Returns(reply);
+        db.ScriptEvaluate(Arg.Any<string>(), Arg.Any<RedisKey[]>(), Arg.Any<RedisValue[]>(), Arg.Any<CommandFlags>())
```
mintycw, line 35: **Minor — test contract gap.** The factory tests verify lease behaviour and fallback, but nothing pins the actual Lua script contract: a regression that swaps `INCRBY`→`INCR`, reshapes the `{count, ttlMs}` return tuple, or reorders `KEYS`/`ARGV` would pass these tests. Add at least one assertion pinning the call shape:

```csharp
await db.Received().ScriptEvaluateAsync(
    Arg.Is<string>(s => s.Contains("INCRBY") && s.Contains("PEXPIRE")),
    Arg.Is<RedisKey[]>(k => k.Length == 1 && k[0] == "rl:clips-write:u:1"),
    Arg.Is<RedisValue[]>(v => v.Length == 2 && (long)v[0] == 1 && (long)v[1] == 60000),
    Arg.Any<CommandFlags>());
```

server/tests/GankedTV.Api.Tests/Services/Caching/RedisRateLimiterFactoryTests.cs
```diff
@@ -0,0 +1,114 @@
+using System.Threading.RateLimiting;
+using FluentAssertions;
+using GankedTV.Api.Services.Caching;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging.Abstractions;
+using NSubstitute;
```
mintycw, line 6: **Minor — follow-up.** All factory tests substitute `IConnectionMultiplexer`; the actual Lua script never executes against Redis in CI. A subtle bug (Lua syntax error, wrong reply parsing, EVAL→EVALSHA caching surprise) would only surface in production. The project already ships a docker-compose Redis 7 — consider a single Testcontainers-driven integration test that runs the limiter against a real Redis, or a `[Fact(Skip = "requires REDIS_URL")]` test wired to env so local verification is one command away. Pinning the script behind a real `EVAL` round-trip is the only way to catch script regressions.

server/tests/GankedTV.Api.Tests/Integration/Endpoints/ClipsReadEndpointsTests.cs
```diff
@@ -374,6 +374,40 @@ public async Task Feed_WithJwt_LikedByMeReflectsLikeRows()
...
+        }
+
+        // The liker hits the same cached page within TTL but must still see likedByMe=true.
+        using (var likerClient = ClientWithBearer(likerToken))
+        {
+            var likerBody = await (await likerClient.GetAsync("/clips/feed")).Content.ReadFromJsonAsync<JsonElement>();
+            likerBody.GetProperty("items").EnumerateArray()
+                .Single(e => e.GetProperty("id").GetGuid() == clipId)
+                .GetProperty("likedByMe").GetBoolean().Should().BeTrue();
+        }
```
mintycw, line 411: **Minor — symmetric coverage.** This test proves `likedByMe` re-stamps after an *anonymous* cache fill — good. The structurally riskier direction isn't covered though: *authed user A populates the cache, authed user B reads it*. Today the leak is structurally impossible because the factory only ever invokes `BuildAnonymousFeedPageAsync`, but a future refactor that accidentally routes an authed caller through a different factory would slip past this test (anonymous always sees `false`, so the assertion is unconditionally satisfied). Add a parallel case where user A likes the clip, hits the feed (warms cache via authed call path), then user B fetches and must see `likedByMe=false` for that clip. Cheap defense-in-depth against an easy-to-introduce regression.

mintycw (commented):
Solid PR with thoughtful architecture — the anonymous-projection split + per-caller `LikedByMe` re-stamping is exactly the right shape for shared-cache correctness, the cluster→in-process limiter degradation matches the "degrade, never fail-open" requirement, and the `RedisOptions` design (lenient URL parse incl. ACL username, no startup crash on bad config) preserves the local-dev story. Coverage is genuinely meaningful for the new code, and the docstrings carry rationale (e.g. why trending isn't invalidated, why INCRBY-before-check) that future maintainers will thank you for.

Main things I'd want addressed before merge — none are data-correctness blockers (the `likedByMe` machinery is sound):

1. Broaden the exception filter in `RedisFixedWindowRateLimiter` so `ObjectDisposedException` / generic `TimeoutException` can't escape as 500s during shutdown races or non-Redis network hiccups.
2. Flow `CancellationToken` into the Redis `ScriptEvaluateAsync` call so client aborts release request resources during Redis slowness.
3. Register `IConnectionMultiplexer` via a factory delegate (not an instance) so the DI container disposes it on `WebApplication.StopAsync` — instance registrations are not disposed by MS.DI.
4. Either invalidate the feed cache on like/unlike or explicitly document the `LikeCount` drift the way trending does (currently the cache docstring reads as if write-time invalidation keeps counts fresh, but likes don't trip it).

The Lua script's INCRBY-before-check is the conventional fixed-window tradeoff and is correctly documented — worth one extra line warning callers about `permitCount > limit` poisoning the bucket. Test-side: factory tests substitute `IConnectionMultiplexer` and never assert the actual `KEYS/ARGV` shape, so a Lua-script regression would pass CI silently — a single contract-pinning assertion would close that gap.

**Verdict: With fixes** — items 1–4 are pre-merge; the rest are Minor / follow-up.

server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
Turbootzz: Done — wrapped the async script eval in `.WaitAsync(cancellationToken)` so a client abort releases the request instead of pinning it until `AsyncTimeout`. `OperationCanceledException` is deliberately not caught — cancellation isn't a Redis failure.

server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
Turbootzz: Done — broadened the degrade filter to a shared `IsRedisFailure` helper covering `RedisException | RedisTimeoutException | ObjectDisposedException | TimeoutException`, used by both the async and sync paths, so a multiplexer disposed during a shutdown/teardown race degrades to in-process instead of 500ing. Cancellation still propagates.

server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
Turbootzz: Done (option a) — added an XML `<remarks>` on `Create` warning callers to use `permitCount <= permitLimit` (the INCRBY-before-check would otherwise poison the window). Every current caller acquires exactly 1 permit, so the over-allocation path is latent; kept the doc over the Lua precheck to stay minimal.

server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
Turbootzz: Done — `Evaluate` now resets `_degradedLogged` on a successful round trip, so a flapping Redis logs once per outage and recovery is observable. A metric/counter is noted as a future observability follow-up.

server/src/GankedTV.Api/Services/Caching/RedisRateLimiterFactory.cs
Turbootzz: Skipping — the built-in DI container doesn''t honor optional/default constructor parameters, so injecting `IConnectionMultiplexer?` when Redis is unconfigured throws at activation (it simply isn''t registered). `IServiceProvider` + `GetService` is the idiomatic way to express this optional dependency, and as you note the per-partition cost is negligible. The factory tests already cover the absent-Redis case via `FactoryWith(null)`.

server/src/GankedTV.Api/Services/Caching/RedisRegistration.cs
Turbootzz: Done — switched to a lazy, container-owned singleton: `AddSingleton<IConnectionMultiplexer>(_ => ConnectionMultiplexer.Connect(configuration))` so the container disposes it on shutdown (clean QUIT), with the cache sharing the same instance via `AddOptions<RedisCacheOptions>().Configure<IServiceProvider>(...)`. Still a single connection, and no nested `BuildServiceProvider()`.

server/src/GankedTV.Api/Services/Caching/RedisRegistration.cs
Turbootzz: Done via the same change — registering a factory (not a pre-built instance) defers `Connect` to first resolve instead of boot, so a slow/unreachable Redis no longer adds a startup penalty. Kept the default `ConnectTimeout` since startup is no longer on that path.

server/src/GankedTV.Api/Services/Caching/FeedCache.cs
Turbootzz: Documented the accepted lag rather than invalidating: likes are high-frequency, so firing `InvalidateFeedsAsync` per like/unlike would defeat the cache. Added a `<para>` to `InvalidateFeedsAsync` noting a cached card''s `LikeCount` is TTL-bounded-stale (same tradeoff trending documents); the detail view and like button reflect the true count immediately.

server/src/GankedTV.Api/Services/Caching/FeedCache.cs
Turbootzz: Skipping — HybridCache already defaults `MaximumPayloadBytes` to 1 MiB, so oversized entries are rejected from L2; feed payloads are bounded (≤100 items, title/description capped by `ClipValidationOptions`) and L1 entries expire in ≤30s. Not load-bearing at current scale, as you noted.

server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
Turbootzz: Skipping — behaviour is identical (you noted this): the key uses the clamped value and the builder re-clamps with the same formula, so they can''t diverge. I''d rather keep clamping owned solely by the builder (single source of truth) than duplicate it at the call site, which is what would actually invite drift.

server/tests/GankedTV.Api.Tests/Services/Caching/RedisRateLimiterFactoryTests.cs
Turbootzz: Done — added an assertion in `Redis_WithinLimit_IsAcquired` pinning the Lua call shape: INCRBY+PEXPIRE present, single key `rl:clips-write:u:1`, ARGV `[1, 60000]`.

server/tests/GankedTV.Api.Tests/Services/Caching/RedisRateLimiterFactoryTests.cs
Turbootzz: Deferring as a follow-up — a Testcontainers-Redis fixture is the right way to pin the live EVAL round-trip but is new test infra disproportionate to this PR. Mitigations: the call-shape is now asserted (sibling thread), and I manually smoke-tested against the compose Redis (30→429 + `Retry-After`, `rl:` key present). Tracking a real-Redis integration test separately.

server/tests/GankedTV.Api.Tests/Integration/Endpoints/ClipsReadEndpointsTests.cs
Turbootzz: Done — added `Feed_CachedFirstPage_FilledByAuthedUser_DoesNotLeakToAnother`: authed user A (who liked the clip) warms the cache, then user B reads the same entry and must see `likedByMe=false`.

mintycw (commented):
One blocker before merge, otherwise this looks great — the fallback guarantees and `likedByMe` re-stamping in particular are very nicely done, and the two-direction cache-leak test (`ClipsReadEndpointsTests.cs:378-444`) is the kind of test I always wish I'd written.

### Must fix

**`Retry-After: 0` on the in-process fallback path** — [`server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs:55-58`](https://github.com/gankedtv/gankedtv/blob/38a3fa5f545855752be6e6166accc49f2cb05ab7/server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs#L55-L58)

```csharp
ctx.HttpContext.Response.Headers.RetryAfter =
    ((int)retryAfter.TotalSeconds).ToString(CultureInfo.InvariantCulture);
```

The Redis path is fine — the Lua script already rounds up to whole seconds before stuffing the metadata. But the in-process `FixedWindowRateLimiter` fallback (Redis down, or `REDIS_URL` unset in dev) returns the *remaining* window time as a sub-second `TimeSpan`. If request 31 arrives at 59.6s into a 60s window, `retryAfter = 400ms` → `(int)0.4 = 0` → `Retry-After: 0`. RFC 7231 says clients can immediately retry on `Retry-After: 0`, so well-behaved clients (and the SPA) tight-loop the limiter.

Mirror the Redis-path math:

```csharp
var seconds = Math.Max(1, (int)Math.Ceiling(retryAfter.TotalSeconds));
ctx.HttpContext.Response.Headers.RetryAfter =
    seconds.ToString(CultureInfo.InvariantCulture);
```

### Nice-to-haves (non-blocking)

- **No test asserts that a throwing `IFeedCache.InvalidateFeedsAsync` doesn't surface as a 500.** The best-effort try/catch in `ClipsMutateEndpoints.cs:206-220` and `MediaJobHostedService.cs:152-158` is the literal codification of the spec's "Redis unavailable → no 500s" criterion, but the existing tests only verify it's *called*. A test that injects `feedCache.InvalidateFeedsAsync(...).Throws<RedisConnectionException>()` and asserts the PATCH/DELETE/MarkReady path still returns 2xx would lock that guarantee in.
- **Cross-pod cache stampede.** `HybridCache.GetOrCreateAsync` single-flights *intra-pod* only. After a `RemoveByTagAsync` clears L2, N pods can each run the feed factory concurrently on the next request. Not a v1 blocker, but worth noting in the PR description as a known scale-up cliff.
- **Shared `RedisRateLimitLease.Acquired` singleton** (`RedisRateLimiterFactory.cs:198`) relies on the base `RateLimitLease.Dispose()` staying a no-op forever. Either allocate per-acquire (cheap) or add a comment pinning the invariant.
- **`redis:7` is unpinned** in `docker-compose.dev.yml` vs the more specific pin on Postgres — minor reproducibility nit.

### Verified, looks correct

- `IReadOnlyList<GameListItem>` / `CachedFeedPage.Items` round-trip through STJ — initially flagged as a serialization risk, but STJ's collection converter handles interface targets and record positional ctors, and `CachedFeedPageSerializationTests` proves it end-to-end. Withdrawn.
- Lua INCRBY+PTTL+PEXPIRE is server-atomic and the "count-before-check is stricter, not looser" reasoning in the comment is right.
- Multiplexer ownership fix (`38a3fa5f`) — DI-disposed singleton with both consumers resolving the same instance is the correct fix.

Once the Retry-After is fixed, this is good to merge.

Turbootzz:
Thanks for the thorough review, @mintycw — all addressed in the latest push (rebased onto current `main`, which pulled in the new staged media pipeline from #113).

**Must fix**
- ✅ **`Retry-After: 0` on the fallback path** — `OnRejected` now does `Math.Max(1, (int)Math.Ceiling(retryAfter.TotalSeconds))`, so the in-process `FixedWindowRateLimiter`'s sub-second remainder can no longer emit `Retry-After: 0`. (Redis path already rounded up; this is a no-op there.)

**Nice-to-haves**
- ✅ **No-500-on-cache-failure tests** — added `Patch_FeedCacheInvalidationThrows_StillReturns200` (endpoint) and `Compress_FeedCacheInvalidationThrows_StillCompletesAndDoesNotFail` (worker). Both inject a throwing `IFeedCache` and assert the committed write/stage still succeeds.
- ✅ **Cross-pod stampede** — documented under "Known limitations (scale-up)" in the PR description.
- ✅ **`RedisRateLimitLease.Acquired` invariant** — added a comment pinning that the shared singleton is safe only while the base `Dispose` stays a no-op.
- ⏭️ **`redis:7` pin** — left as-is: it matches the `postgres:18` major-pin convention already used here (`minio/minio` is the only actually-untagged image, and it's pre-existing/out of scope).

Note on the rebase: `#113` replaced `MediaJobHostedService` with the `ThumbnailWorker`/`CompressWorker` stages, so the "new clip → invalidate feed" hook was re-homed into those workers (best-effort, fires when a clip reaches `ready`) via a shared `ClipStageWorker.InvalidateFeedsBestEffortAsync`.

mintycw (approved):
Reviewed via /superpowers:requesting-code-review.

Core correctness contracts hold: no cross-user leak (anonymous projection + per-caller `likedByMe` re-stamp, pinned by integration tests), fail-back never fail-open (conservative `IsRedisFailure` + local `FixedWindowRateLimiter` fallback), mutations never 500 on cache failure (best-effort invalidation tests). Lua script is atomic (single EVAL with INCRBY + PTTL + conditional PEXPIRE, with `PTTL < 0` guard). Multiplexer lifecycle correct (DI singleton, shared between L2 + limiter). `Retry-After` floored at 1s to defend against RFC 7231 "retry immediately on 0".

Non-blocking follow-ups worth tracking:
- View-policy rate limit is per-pod by design — either extend to Redis or comment the asymmetry (`ClipsRateLimiting.AddClipsViewPolicy`).
- No recovery-path test (fallback engagement tested, disengagement after Redis recovers is not; `_degradedLogged` re-arm untested).
- No test resolves `AddGankedCaching` without Redis configured — a future ctor dep on `IConnectionMultiplexer` would only fail at first request in prod.
- No real Redis-path round-trip through HybridCache — `FeedCacheTests` is L1-only and standalone STJ tests don't exercise HybridCache's L2 codec. A testcontainers test would close it.
- Hoist `RejectionStatusCode = 429` + `OnRejected` into a shared `AddSharedRateLimiterDefaults` to remove the implicit ordering dependency between `AuthRateLimiting` and `ClipsRateLimiting`.
- Version-prefix cache keys (`feed:v1:...`) to prevent silent collisions on future schema changes.
- Trending `since` is frozen into the cached page — fine for 24h/7d, add a comment that TTL must stay « window.

PR #113, opened 2026-05-23
[server][web] Compress clips on upload + just-in-time adaptive playback

### LINKED ISSUE #102: [server][web] Video transcoding: HLS multi-quality pipeline + adaptive playback
## Summary

Phase 4 (Scale & Grow) — transcode uploaded clips into adaptive HLS so playback works on slow connections and mobile. Today the media worker only extracts a thumbnail and the player streams the **single original MP4** via a presigned URL ([ClipView.vue](web/src/views/ClipView.vue) sets `el.src = detail.videoUrl` on a plain `<video>` driven by Plyr). There is no quality ladder and no adaptive streaming.

Per [PLAN.md §9](PLAN.md) Phase 4: *Video transcoding (multiple qualities with FFmpeg)*.

Builds directly on the existing DB-backed media job pipeline — no hard dependency on other Phase 4 issues.

## Server

- Extend the media pipeline to produce an HLS rendition ladder + `master.m3u8`. The worker ([MediaJobHostedService.cs](server/src/GankedTV.Api/Services/Media/MediaJobHostedService.cs)) already leases `processing` clips and runs ffmpeg via [FfmpegRunner.cs](server/src/GankedTV.Api/Services/Media/FfmpegRunner.cs); add a `TranscodeJobService` alongside [ThumbnailJobService.cs](server/src/GankedTV.Api/Services/Media/ThumbnailJobService.cs).
- **Ladder:** e.g. 1080p / 720p / 480p, **source-capped** (never upscale). New knobs on [MediaJobOptions.cs](server/src/GankedTV.Api/Services/Media/MediaJobOptions.cs) (ladder definition, segment duration, enable flag).
- **Status flow:** add `transcoding` between `processing` and `ready` (extend `ClipStatuses` + the lease/poll query). A failed transcode lands in `status='failed'` and must **not wedge the worker** (respects `MaxAttempts`).
- **Storage:** write segments + playlists under a key prefix (e.g. `hls/{clipId}/...`) in the clips bucket, or a new `hls` bucket bootstrapped in [BucketBootstrapHostedService.cs](server/src/GankedTV.Api/Services/ObjectStorage/BucketBootstrapHostedService.cs). Upload via [S3ObjectStorageService.cs](server/src/GankedTV.Api/Services/ObjectStorage/S3ObjectStorageService.cs).
- **Clip detail:** add `hlsUrl` (presigned master playlist) alongside the existing `videoUrl` (MP4 **fallback** — keep the original). Update [ClipMappings.cs](server/src/GankedTV.Api/Contracts/Clips/ClipMappings.cs) + the detail DTO.
- Backfill consideration: existing clips have no HLS — `hlsUrl` is null for them and the player falls back to MP4. A re-enqueue/backfill path is optional follow-up.

## Web

- Add `hls.js`; in [ClipView.vue](web/src/views/ClipView.vue) load `hlsUrl` through hls.js + Plyr when present, falling back to `videoUrl` (plain MP4) otherwise. Use **native HLS on Safari** (no hls.js).
- Quality selector wired into Plyr's controls (`settings: ['quality']`), reading levels from the manifest.

## How to test manually

```bash
# Upload a clip, then watch the worker produce HLS
curl -X POST http://localhost:5050/clips/<id>/complete -i
# poll status: processing → transcoding → ready
curl -s http://localhost:5050/clips/<id> | jq '{status, hlsUrl, videoUrl}'
# segments present in storage (MinIO console / mc ls)

# Browser
# Open the clip → quality menu appears, switching renditions works
# An older clip (no HLS) → still plays via MP4 fallback
```

## Checklist

- [ ] Upload produces a rendition ladder + `master.m3u8` in storage (source-capped, no upscaling)
- [ ] Detail returns presigned `hlsUrl`; MP4 `videoUrl` retained as fallback
- [ ] Player switches quality; native HLS used on Safari
- [ ] Clips without HLS still play (MP4 fallback)
- [ ] Failed transcode → `status='failed'`, worker continues (no wedge), respects `MaxAttempts`

Turbootzz:
Closes #102

## What & why

Reworks the media pipeline from "stream the single original MP4" into a **disk-saving, Tdarr-style** model: each clip is re-encoded once into a single compressed master (the original is deleted), and adaptive HLS ladders are generated **just-in-time at watch time** and cached transiently — never stored permanently. Net persistent storage per clip goes *down*, while slow/mobile playback still gets adaptive quality.

> Note: this evolved during review from an earlier "persist a 1080/720/480 HLS ladder alongside the original" approach — which *increased* storage — to the compression-first design here. The diff reflects the final design.

## Pipeline

Three stages on a shared, DB-queue-backed worker (`MediaStageWorker<TJob>`, `FOR UPDATE SKIP LOCKED`):

1. **Thumbnail** (`processing`) — poster + ffprobe metadata, then advances to `transcoding`.
2. **Compress** (`transcoding`) — re-encodes the upload into one resolution-capped, quality-targeted master (H.264 in dev, AV1 on a GPU host), repoints `video_key`, records `video_codec`, **deletes the original**, advances to `ready`.
3. **JIT stream** (`clip_stream_jobs`, watch-time) — for devices that can't decode the master (e.g. AV1), `GET /clips/{id}/stream` builds an H.264 HLS ladder on demand into the public-read **`stream-cache`** bucket (auto-evicted by a lifecycle rule); the player polls until ready. Because clips are short, whole-clip JIT is fast — no segment-level complexity.

The web player plays the master directly when the browser can decode it (H.264 always; AV1 on capable devices), otherwise falls back to the JIT H.264 stream (hls.js + Plyr quality menu).

## Production GPU split

The compress + JIT stages are GPU work and are location-independent (DB queue). Env toggles let them run on a separate host (e.g. TrueNAS + NVENC) while the main API runs only the thumbnail stage — see the table in `CLAUDE.md`. Master encoder, JIT encoder, resolution cap, and CRF are all configurable, so moving to AV1 is config, not code.

## Notable decisions / trade-offs

- **Irreversible by design** (Tdarr-style): deleting the original means re-encodes come from a lossy master.
- **`stream-cache` is public-read** with unguessable UUID keys, mirroring the existing `game-covers` bucket (clips are public/unlisted only).
- A fresh upload shows a **"Processing…"** state and polls instead of bouncing to not-found while the longer pipeline runs.
- Client-side **poster preview** on file pick so the upload form isn't empty while filling in details.

## Testing

- Server: 713 tests, coverage 95% line / 88% branch (gate 85/85).
- Web: type-check + lint clean, 169 tests.
- Real `ffmpeg` validated for both the compress and JIT-ladder commands; live dev smoke test confirmed bucket bootstrap + lifecycle on MinIO and a full upload → compress → `ready` run (original deleted, single master stored).

## Migrations

One squashed migration `AddVideoCompressionAndStreamJobs` (adds `clips.video_codec`, the `clip_stream_jobs` table + indexes; drops the interim `hls_key`). Requires `dotnet ef database update` / `make migrate`.

## Release Notes

* **New Features**
  * Added two-stage video processing pipeline with compression and just-in-time H.264 transcoding for codec compatibility
  * Introduced new `/stream` endpoint for HLS playback with adaptive quality selection
  * Added processing status indicator during clip transcoding
  * Implemented poster image generation from uploaded videos
  * Enabled AV1 codec support with automatic H.264 fallback for incompatible devices

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/113?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Turbootzz:
@coderabbitai review

Turbootzz:
@coderabbitai review

Turbootzz:
@coderabbitai review

server/src/GankedTV.Api/Services/Media/ClipStreamJobStore.cs
```diff
@@ -0,0 +1,121 @@
...
+        var clip = await _db.Clips
+            .AsNoTracking()
+            .Where(c => c.Id == job.ClipId)
+            .Select(c => new { c.VideoKey, c.Height, c.Status })
+            .FirstOrDefaultAsync(ct);
+
+        if (clip is null || clip.Status != ClipStatuses.Ready)
+        {
+            // Source gone / not ready — drop the stale job so it doesn't churn the queue.
+            await _db.ClipStreamJobs.Where(j => j.ClipId == job.ClipId).ExecuteDeleteAsync(ct);
```
mintycw, line 89: **Important:** dropping a stale stream-job row here without invalidating any already-cached `stream-cache/{clipId:N}/*` objects means a clip that was once `ready` and got a JIT ladder, then was deleted or flipped to `failed`, still has a publicly readable HLS rendition for up to `StreamCacheTtlDays` (default 14). For unlisted clips this is an unbounded exposure window keyed on a guessable-once-leaked clip GUID. Either delete the cached prefix here (and in `ClipBlobCleanup.TryDeleteAsync`) or document the residual surface explicitly.

server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
```diff
@@ -35,10 +36,45 @@ public static IEndpointRouteBuilder MapClipsReadEndpoints(this IEndpointRouteBui
...
+
+        if (await streamJobs.GetStatusAsync(id, ct) == ClipStreamJobStatuses.Failed)
+            return Results.Problem(
+                statusCode: StatusCodes.Status503ServiceUnavailable,
+                title: "stream_unavailable",
+                detail: "Just-in-time transcode failed for this clip.");
+
+        await streamJobs.EnqueueAsync(id, ct);
+        return Results.Json(new StreamResponse(null, "pending"), statusCode: StatusCodes.Status202Accepted);
+    }
```
mintycw, line 92: **Important:** `GET /clips/{id}/stream` is anonymous and has no `RequireRateLimiting(...)` attached, unlike the sibling view endpoint (`ClipsViewPolicy`). The enqueue is idempotent (ON CONFLICT DO NOTHING), but each call still performs a DB lookup + S3 HEAD + a second DB roundtrip on miss. A trivial loop from an unauth client multiplies that per second per clip with no upper bound. Consider attaching `.RequireRateLimiting(ClipsRateLimiting.ClipsViewPolicy)` or a dedicated stream policy.

mintycw (commented):
## Review summary

Solid, well-factored implementation of the Tdarr-style pipeline. The shared `MediaStageWorker<TJob>` base nicely captures the lease/retry/drain invariants, and the tests exercise the three concrete subclasses against it. The Tdarr trade-off (delete original, JIT for incompatible playback) is clearly intentional, and the doc comments are unusually good — they call out the orderings (DB advance THEN delete; segments uploaded BEFORE master.m3u8) that matter for correctness. The codec-driven web player + Plyr + hls.js integration is clean, including the `playerToken` cancellation pattern and the silent-source ffprobe fallback in `JitLadderService`.

### Strengths
- `MediaStageWorker<TJob>` cleanly factors the loop / probe / drain-cap / fresh-token-on-fail invariants — adding a fourth stage in the future is a small subclass.
- Claim queries use `FOR UPDATE SKIP LOCKED` inside an explicit transaction with `AsNoTracking()`, and every state mutation guards on `expectedAttempt` + status, so a stale lease never clobbers a re-claimed row.
- ffmpeg stderr is run through `RedactUrls` everywhere before being embedded in exceptions/logs — the presigned-URL signature can't leak via an error envelope.
- Compress writes the new master to a distinct key (`.cmp.mp4`) and the worker only deletes the original AFTER the DB is repointed; even a crash between encode and DB-update is recoverable.
- JIT uploads segments first and `master.m3u8` last, so the `/stream` endpoint's "master exists = ready" probe can never expose a playlist pointing at missing segments.
- Good test coverage of the cross-cutting behaviors (drain cap, startup probe, attempt-guarded fail/release), with integration tests against real Postgres for the queue stores.

### Main themes from the review (also inline)

- **JIT cache lifetime is decoupled from the clip lifecycle.** Deleting a clip (or flipping it from public to unlisted) does not invalidate the public-read `stream-cache/{clipId:N}/master.m3u8` + segment objects — they live until the 14-day lifecycle rule fires. For unlisted clips whose ID later leaks this is a real exposure window. **`ClipBlobCleanup.TryDeleteAsync` (file unchanged in this PR, so noted here) should be extended to delete the cached HLS prefix** alongside the video + thumbnail.
- **`/clips/{id}/stream` is unauthenticated and unrate-limited.** Idempotent on the DB side (ON CONFLICT DO NOTHING), but a flood still drives a DB roundtrip + an S3 HEAD per request, and once a clip lands in `failed` state there is no recovery path short of a manual row delete.
- **Two compress-worker race on lease expiry.** With `LeaseDuration=5min` and `TranscodeTimeout=10min`, a stuck-but-alive Worker A can encode in parallel with the re-claimed Worker B; both write the same deterministic `.cmp.mp4` key. The DB outcome is guarded by `expectedAttempt`, but the S3 object can briefly flap and Worker A can delete the original after Worker B has already pointed the clip at the (now-overwritten) master.
- A small set of polish items inline.

### Issues raised inline
- Critical: 0
- Important: 6 (including the JIT-cache-on-delete issue called out above)
- Minor: 5

### Recommendations
- Extend `ClipBlobCleanup` to remove the `stream-cache/{clipId:N}/` prefix on clip delete (and consider on visibility flip from public→unlisted, or accept the residual exposure window explicitly in a comment).
- Either gate `/clips/{id}/stream` behind the existing view rate-limit policy, or document explicitly that the JIT enqueue is intended to be unbounded.
- Raise `LeaseDuration` above `TranscodeTimeout` (or use a per-stage lease) so a still-running encode can never have its row re-claimed.
- Treat the no-recovery-from-`failed` state for `clip_stream_jobs` as a follow-up: today a flaky GPU box can permanently poison playback for a clip.

server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
```diff
@@ -35,10 +36,45 @@ public static IEndpointRouteBuilder MapClipsReadEndpoints(this IEndpointRouteBui
...
+        {
+            var url = S3PublicUrls.BuildUrl(s3.Value, s3.Value.StreamCacheBucket, masterKey);
+            return Results.Ok(new StreamResponse(url, "ready"));
+        }
+
+        if (await streamJobs.GetStatusAsync(id, ct) == ClipStreamJobStatuses.Failed)
+            return Results.Problem(
+                statusCode: StatusCodes.Status503ServiceUnavailable,
+                title: "stream_unavailable",
+                detail: "Just-in-time transcode failed for this clip.");
```
mintycw, line 89: **Important:** once a `clip_stream_jobs` row lands in `failed`, there is no recovery path. `EnqueueAsync` is `ON CONFLICT DO NOTHING` so subsequent `/stream` calls keep finding the `failed` row and 503ing forever, even if the GPU box is fixed an hour later. Consider one of: (a) `MarkFailedAsync` deletes the row instead of setting status='failed' (the cache miss naturally re-enqueues), (b) `EnqueueAsync` resets `failed` rows back to `pending` with `processing_attempts=0` on conflict, or (c) a maintenance sweep that retries failed rows after some cooldown. The current behavior turns a transient GPU outage into a permanent stream blocker per clip.

server/src/GankedTV.Api/Services/Media/CompressJobService.cs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    // The compressed master lives at a distinct key (…/{clipId}.cmp.mp4) so the encode never
+    // overwrites the source mid-job; the worker deletes the original only after the DB has
+    // been repointed at this key.
+    internal static string CompressedKeyFor(string originalKey) =>
+        originalKey.EndsWith(".mp4", StringComparison.OrdinalIgnoreCase)
+            ? originalKey[..^4] + ".cmp.mp4"
+            : originalKey + ".cmp.mp4";
```
mintycw, line 80: **Important:** `CompressedKeyFor` is deterministic per `originalKey`, which combined with a lease/timeout mismatch creates a race. With defaults `LeaseDuration=5min` and `TranscodeTimeout=10min`, a slow-but-not-yet-failed Worker A can have its row re-claimed by Worker B at the 5-minute mark; both then encode in parallel and both `PutObjectAsync` to the same `{clipId}.cmp.mp4` key. The DB advance is guarded by `expectedAttempt`, but the S3 object can flap (B finishes first → clip points at it → A overwrites it), and the `CompressWorker` finally branch then deletes the original even for the loser. Two safer options: (1) make `LeaseDuration > TranscodeTimeout` (or a per-stage lease) so an in-flight encode is never re-claimed, or (2) include attempt number in the output key so the two encodes can't collide.

server/src/GankedTV.Api/Services/Media/MediaJobOptions.cs
```diff
@@ -37,4 +37,73 @@ public sealed class MediaJobOptions
...
+    // Whether this instance runs the thumbnail BackgroundService (claims 'processing').
+    public bool ThumbnailWorkerEnabled { get; set; } = true;
+
+    // Whether this instance runs the GPU workers — the compress stage (claims 'transcoding')
+    // and the JIT stream-rendition stage (claims pending clip_stream_jobs).
+    public bool TranscodeWorkerEnabled { get; set; } = true;
+
+    // Hard ceiling on a single ffmpeg encode (compress or a JIT ladder). Far larger than
+    // ProcessTimeout because a full encode takes much longer than a one-frame thumbnail.
+    public TimeSpan TranscodeTimeout { get; set; } = TimeSpan.FromMinutes(10);
```
mintycw, line 63: **Important:** `LeaseDuration` (5min default) is well below `TranscodeTimeout` (10min default). For the thumbnail stage where `ProcessTimeout=2min` this is correct, but for both compress and JIT the encode can legitimately run longer than the lease, which lets another worker re-claim the row mid-encode (see comment on `CompressedKeyFor`). Either invert these defaults (lease > transcode timeout) or split into a `ThumbnailLeaseDuration` + `TranscodeLeaseDuration` so each stage's lease comfortably exceeds its own timeout.

web/src/views/ClipView.vue
```diff
@@ -149,12 +202,104 @@ function attachViewTracking(targetClipId: string, el: HTMLVideoElement) {
...
+    if (myToken !== playerToken) return
+    if (res.status === 'ready' && res.hlsUrl) {
+      attachHlsStream(el, res.hlsUrl)
+      return
+    }
+    await new Promise((r) => setTimeout(r, JIT_POLL_MS))
+    if (myToken !== playerToken) return
+  }
+  // Exhausted the poll budget without a ready rendition — surface an error rather than hang.
+  errored.value = true
```
mintycw, line 259: **Important:** the JIT poll budget exhausts silently into `errored.value = true` with a generic "Couldn't load this clip" panel. From the user's view (just hit play on what looked like a perfectly fine thumbnail) this is indistinguishable from a 500 on the detail endpoint. Given that the JIT failure mode is meaningfully different ("this clip needs a transcode that hasn't finished or has failed"), worth a specific message — and ideally a Retry that re-enters `startJitStream` rather than re-running `loadClip` (which only re-fetches detail and re-decides codec; the JIT failure isn't surfaced through that path).

server/src/GankedTV.Api/Services/Media/ClipStreamJobStore.cs
```diff
@@ -0,0 +1,121 @@
...
+                FROM clip_stream_jobs
+                WHERE status = 'pending'
+                  AND (processing_started_at IS NULL OR processing_started_at < {leaseExpiry})
+                  AND processing_attempts < {maxAttempts}
+                ORDER BY updated_at
+                FOR UPDATE SKIP LOCKED
+                LIMIT 1
+            ")
+            .AsNoTracking()
+            .ToListAsync(ct);
```
mintycw, line 69: **Minor:** `FOR UPDATE SKIP LOCKED ... LIMIT 1` materialized via `.ToListAsync()` + `AsNoTracking()` is fine for correctness (the lock is held by the surrounding transaction), but the follow-up `ExecuteUpdateAsync` runs in the same transaction and issues a second statement that re-resolves the row by PK; on a contended queue this is two roundtrips instead of one. The thumbnail-stage `ClipMediaJobStore.ClaimNextAsync` uses the identical pattern and has a comment explaining the choice — worth mirroring that comment here so the next reader doesn't try to "optimize" by switching to `UPDATE ... WHERE id = (SELECT ... FOR UPDATE SKIP LOCKED)`.

server/src/GankedTV.Api/Services/Media/JitLadderService.cs
```diff
@@ -0,0 +1,259 @@
...
+                throw new InvalidOperationException("ffmpeg JIT transcode produced no master.m3u8");
+            }
+
+            var keyPrefix = BuildCachePrefix(job.ClipId);
+            // Upload segments + variant playlists first and master.m3u8 LAST. The /stream
+            // endpoint treats the presence of master.m3u8 as "ready", so publishing it before
+            // its referenced files exist would briefly serve a playlist whose segments 404.
+            const string masterName = "master.m3u8";
+            var paths = Directory.EnumerateFiles(workDir)
+                .OrderBy(p => Path.GetFileName(p) == masterName) // false (0) sorts before true (1)
```
mintycw, line 74: **Minor:** sorting so that `master.m3u8` lands last keeps the endpoint's "master.m3u8 exists = ready" invariant, but the ordering of segment writes themselves is `Directory.EnumerateFiles` order — implementation-defined. In practice ffmpeg flushes segments before the variant playlists, so this is fine today, but consider explicitly ordering by extension (`.ts`/`.m4s` first, then `.m3u8`, then `master.m3u8`) to make the contract self-evident and survive a future ffmpeg version that flushes variant playlists earlier.

server/src/GankedTV.Api/Services/Media/JitLadderService.cs
```diff
@@ -0,0 +1,259 @@
...
+        }
+        finally
+        {
+            TryDeleteDir(workDir);
+        }
+    }
+
+    // Object-key prefix for one clip's cached HLS output. Relative URIs inside the playlists
+    // resolve against this prefix, so master + variant playlists + segments all live here.
+    internal static string BuildCachePrefix(Guid clipId) => $"{clipId:N}";
```
mintycw, line 97: **Minor:** the cache prefix is purely the clip GUID, which is exposed in every clip-detail response. Combined with the public-read `stream-cache` bucket, anyone who has seen the clip detail page (including via shareCode for an unlisted clip) can construct the HLS URL directly, bypassing the `/stream` endpoint entirely. That's broadly consistent with how presigned-master URLs work today, but the JIT cache has a 14-day TTL versus the 1-hour presign window — so the effective "share window" for an unlisted clip's content is materially longer. Consider prefixing with a random opaque token stored on the job row, or just document this trade-off.

server/src/GankedTV.Api/Services/Media/CompressWorker.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            var storage = scope.GetRequiredService<IObjectStorageService>();
+            var clips = scope.GetRequiredService<IOptionsMonitor<S3Options>>().CurrentValue.ClipsBucket;
+            try
+            {
+                await storage.DeleteObjectAsync(clips, result.OriginalKey, ct);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning(ex,
```
mintycw: **Minor:** the comment correctly notes the DB-advance-then-delete ordering, but on a transient S3 delete error the orphaned original silently consumes the very disk savings the pipeline exists to deliver. Consider logging at `Error` (not `Warning`) so this is actually paged on / dashboarded, or enqueuing a maintenance-sweep entry to retry the delete later — otherwise these orphans only get noticed when someone audits bucket size by hand.

server/src/GankedTV.Api/Services/ObjectStorage/S3ObjectStorageService.cs
```diff
@@ -46,23 +46,71 @@ public async Task EnsureBucketsAsync(CancellationToken ct = default)
...
         {
             _logger.LogWarning(
-                "GameCoversBucket '{Bucket}' aliases a private bucket; skipping the anonymous-read "
-                + "policy to avoid exposing clips/thumbnails.", _options.GameCoversBucket);
+                "{OptionName} '{Bucket}' aliases a private bucket; skipping the anonymous-read "
+                + "policy to avoid exposing clips/thumbnails.", optionName, bucket);
             return;
         }
 
         await _s3.PutBucketPolicyAsync(new PutBucketPolicyRequest
```
mintycw, line 110: **Minor:** `PutBucketPolicyAsync` isn't wrapped in a try/catch the way `PutLifecycleConfigurationAsync` is. If a future S3-compatible backend doesn't support `s3:PutBucketPolicy` (or the credentials lack permission in a hardened prod setup), startup aborts via `BucketBootstrapHostedService.LogCritical + throw`. Lifecycle has the right shape; the policy call probably wants the same warn-and-continue treatment, or a clear comment that this MUST succeed (otherwise game covers / JIT URLs are unreachable).

server/src/GankedTV.Api/Services/Media/ClipStreamJobStore.cs
Turbootzz: Applied. `ClipBlobCleanup.TryDeleteAsync` now clears the `stream-cache/{clipId}/` prefix (new `DeleteByPrefixAsync`) when a clip is deleted, so a deleted clip's cached HLS no longer lingers until the lifecycle TTL. The "flipped to `failed`" sub-case doesn't arise: a clip only gets a cached ladder after reaching `ready`, and clips never transition `ready → failed`.

server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
Turbootzz: Applied. `/clips/{id}/stream` now has `.RequireRateLimiting(ClipsRateLimiting.ClipsViewPolicy)` — the same per-IP bucket as the sibling view endpoint.

server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
Turbootzz: Applied. `EnqueueAsync` now resets a `failed` row back to `pending` once it is older than a 5-minute cooldown (conditional `ON CONFLICT DO UPDATE ... WHERE status='failed' AND updated_at < cutoff`). The endpoint enqueues *before* the failed-check, so a transient GPU outage recovers on the next request, while a permanently-broken clip still only retries periodically (bounded), avoiding a thrash loop.

server/src/GankedTV.Api/Services/Media/CompressJobService.cs
Turbootzz: Applied (option 1). `LeaseDuration` default is now 15 min, comfortably above `TranscodeTimeout` (10 min), so an in-flight compress/JIT encode is never re-claimed mid-run — this removes the parallel-encode / same-key object-flap race entirely (preferred over encoding the attempt number into the key, which would only de-collide the output rather than prevent the wasted second encode).

server/src/GankedTV.Api/Services/Media/MediaJobOptions.cs
Turbootzz: Applied — same root cause as the `CompressedKeyFor` race. Fixed by making `LeaseDuration` (now 15 min) exceed `TranscodeTimeout` (10 min). Kept a single shared lease rather than splitting per-stage: a 15-min lease is harmless for the seconds-long thumbnail stage (it only delays re-claim of a *crashed* thumbnail job), and avoids adding a second knob.

web/src/views/ClipView.vue
Turbootzz: Applied. JIT failures (poll-budget exhaustion, unsupported playback, or a fetch error) now surface a distinct message — "This clip is still being prepared for your device — try again in a moment." — via a dedicated `errorMessage`, instead of the generic detail-load copy. The existing Retry button re-runs `loadClip`, which tears down the player and re-enters `setupPlayer → startJitStream`, so it is a genuine fresh JIT attempt rather than just a detail re-fetch.

server/src/GankedTV.Api/Services/Media/ClipStreamJobStore.cs
Turbootzz: Applied — added the explanatory comment mirroring `ClipMediaJobStore.ClaimNextAsync` (lock held by the surrounding transaction; two statements kept deliberately; don't fold into a single `UPDATE ... WHERE id = (SELECT ... FOR UPDATE)`).

server/src/GankedTV.Api/Services/Media/JitLadderService.cs
Turbootzz: Not changing — intentional. `master.m3u8` is the only externally-observable entrypoint: it is both the `/stream` readiness signal and the only URL handed to clients. It is now uploaded last, so variant playlists/segments are never fetched until it exists; their relative upload order among themselves is invisible to any client. Explicit `.ts`-then-`.m3u8` ordering would add code to guarantee something nothing observes.

server/src/GankedTV.Api/Services/Media/JitLadderService.cs
Turbootzz: Keeping the clip-GUID prefix (documented in `S3Options.StreamCacheBucket`). This matches the already-approved public-`stream-cache` design — the same posture as the public `game-covers` bucket, and the clip GUID / shareCode is already in the detail response. The concrete risk you raised (a *deleted* clip's HLS lingering for the TTL) is now closed by clearing the cache prefix on deletion. For a live unlisted clip, the content is link-shareable by design, so an opaque token would add a job-row column + URL plumbing for a marginal gain in this threat model. Happy to revisit if unlisted is meant to be a stronger guarantee.

server/src/GankedTV.Api/Services/Media/CompressWorker.cs
Turbootzz: Applied — the orphaned-original delete failure now logs at `Error` (was `Warning`) so it's alertable/dashboardable. A maintenance-sweep retry is a reasonable follow-up but felt out of scope for this PR.

server/src/GankedTV.Api/Services/ObjectStorage/S3ObjectStorageService.cs
Turbootzz: Intentional asymmetry, not an oversight. The lifecycle rule is an optimization (auto-eviction), so it warns-and-continues. The bucket *policy* is load-bearing: without anonymous-read, game covers and JIT HLS URLs are unreachable, so failing fast at startup is the correct behavior rather than silently booting into broken public media. Leaving the policy call fatal by design.

mintycw (approved):
Reviewed via /superpowers:requesting-code-review. No must-fix blockers — pipeline architecture is sound (delete-after-DB-repoint sequencing is correctly transactional, MediaStageWorker<TJob> abstraction supports the GPU-split deployment story, lease/attempt guards prevent double-write).

Non-blocking follow-ups worth tracking (not gating merge):
- Viewer-race double-encode window on JIT cache-miss (wastes GPU, not data loss) — persist master_key on the job row to use DB as readiness oracle
- Add startup validator: `LeaseDuration > TranscodeTimeout` (the comment in MediaJobOptions warns about it; enforce it)
- Tie CompressJobService presigned-input TTL to TranscodeTimeout so bumping the timeout past 30min can't expire the URL mid-encode
- Distinguish fresh-failed (within 5-min cooldown) from pending in the /stream 503 copy
- Confirm `make seed` only produces H.264-compatible videos (pre-existing ready clips have video_codec=null and player optimistically plays master)

PR #114, opened 2026-05-24
[server][web] Game leaderboards: most-liked per game per week + /leaderboards

### LINKED ISSUE #104: [server][web] Game leaderboards: most-liked per game per week + /leaderboards
## Summary

Phase 4 (Scale & Grow) — per-game leaderboards ranking the most-liked clips in a time window. Today [TrendingView.vue](web/src/views/TrendingView.vue) renders a single global top-10 sorted in-memory by `likeCount`; there's no per-game ranking and no windowed aggregation in the schema or query layer.

Per [PLAN.md §9](PLAN.md) Phase 4: *Game leaderboards (most liked per game per week)*.

Soft synergy with #87 (windowed like/view counts) and can use #1's Redis cache when present. No hard dependency.

## Server

- `GET /games/{slug}/leaderboard?window=week|month|all` — top clips for one game, ranked by likes accrued within the window. Extends [GamesEndpoints.cs](server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs).
- `GET /leaderboards?window=week` — global view: top games (by windowed likes) + top clips overall.
- Aggregate over [Like.cs](server/src/GankedTV.Api/Data/Entities/Like.cs)`.CreatedAt` within the window (not the denormalized `clip.like_count`, which is all-time), joined to clips/games. Add a supporting index on `likes(created_at)` if `EXPLAIN` shows a seq scan.
- Filter to `public` + `ready` clips. Cache results via #1's Redis when available (short TTL); fine to compute on read otherwise.
- DTOs in `Contracts/Leaderboards/` with a `LeaderboardEntry` (rank + clip summary + windowed like count) + hand-written mapper, per the repo's mapper convention.

## Web

- Add `web/src/api/leaderboard.ts` (`game(slug, {window})`, `global({window})`).
- New `/leaderboards` route + `LeaderboardView.vue` ([web/src/router/index.ts](web/src/router/index.ts)); ranked list reusing [ClipCard.vue](web/src/components/ClipCard.vue) with a rank badge; window tabs (week/month/all).
- Surface a "Top this week" leaderboard block on [GameView.vue](web/src/views/GameView.vue).
- Add a "Leaderboards" nav link in [AppNav.vue](web/src/components/AppNav.vue).

## How to test manually

```bash
curl -s 'http://localhost:5050/leaderboards?window=week' | jq
curl -s 'http://localhost:5050/games/<slug>/leaderboard?window=week' | jq '.entries[] | {rank, title, likes}'
# Browser: /leaderboards toggles window; /game/<slug> shows a top-this-week block
```

## Checklist

- [ ] Ranking counts only likes within the window (boundary respected; uses `likes.created_at`, not all-time `like_count`)
- [ ] Ties ranked deterministically and stably
- [ ] Only `public` + `ready` clips appear
- [ ] Unknown `window` value returns 400; empty window renders gracefully

mintycw:
## Summary

Per-game and global leaderboards ranking clips by likes accrued within a time window (week/month/all), rather than the denormalized all-time `clip.like_count`. Surfaces a standalone `/leaderboards` page and a "Top This Week" block on each game page.

## What's here

- **Server:** `GET /games/{slug}/leaderboard` and `GET /leaderboards` (top clips + top games). Aggregates over `likes.created_at`, filters to `public`+`ready`, deterministic tiebreak (likes desc, createdAt desc, id asc). Contracts/Leaderboards/ with hand-written mapper. Window parser kept separate from trending's so the two endpoints can evolve their window vocabularies independently.
- **Web:** `api/leaderboards.ts` (`global` + `forGame`), `/leaderboards` route + view (window tabs, top clips + top games sidebar), `GameLeaderboardBlock` embedded on `GameView` that self-hides when there's no activity in the window, shared `LeaderboardRow` component, Leaderboards link in `AppNav`.
- **Tests:** 13 server integration tests covering window boundaries, public/ready filtering, deterministic+stable ties, 400/404 paths, empty windows, per-game scoping. Vitest specs for the api module.

Closes #104

## How to test manually

API watcher needs a restart after this lands — new endpoint registrations can't hot-reload. From the worktree (`make ports` for the local URLs):

```bash
curl -s 'http://localhost:5050/leaderboards?window=week' | jq
curl -s 'http://localhost:5050/games/valorant/leaderboard?window=week' \
  | jq '.entries[] | {rank, title: .clip.title, windowLikes}'
curl -s -o /dev/null -w '%{http_code}\n' 'http://localhost:5050/leaderboards?window=bogus'     # 400
curl -s -o /dev/null -w '%{http_code}\n' 'http://localhost:5050/games/no-such/leaderboard?window=week'  # 404
```

Browser: top-nav **Leaderboards** toggles the week/month/all tabs and refetches; `/game/<slug>` shows a "Top This Week" block above the clip grid that disappears when the game has no in-window likes.

## Checklist

- [ ] Verified manually

## Release Notes

* **New Features**
  * Added a global leaderboards page displaying top-ranked clips and games filtered by time period
  * Implemented time-window filtering for leaderboards with week, month, and all-time options
  * Added a game-specific leaderboard section on individual game pages showing top clips for that game
  * Added navigation menu link to access the new leaderboards feature

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/114?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/src/GankedTV.Api/Contracts/Leaderboards/LeaderboardWindow.cs
+++ b/server/src/GankedTV.Api/Contracts/Leaderboards/LeaderboardWindow.cs
@@ -1,0 +1,35 @@
+namespace GankedTV.Api.Contracts.Leaderboards;
+
+// Leaderboards intentionally use coarser windows than trending (week/month/all vs 24h/7d):
+// "most liked this hour" is too noisy to rank, and "all-time" is a meaningful question for
+// boards but never for trending. Kept separate from TryParseTrendingWindow so the two
+// endpoints can evolve their window vocabularies independently.
+public static class LeaderboardWindow
+{
+    public const string Week = "week";
+    public const string Month = "month";
+    public const string All = "all";
+    public const string Default = Week;
+
+    // `since` for `all` is DateTimeOffset.MinValue so the same `l.CreatedAt >= since`
+    // filter covers every window with no branching at the query layer.
+    public static bool TryParse(string? window, out DateTimeOffset since)
+    {
+        var now = DateTimeOffset.UtcNow;
+        switch (window)
+        {
+            case Week:
+                since = now.AddDays(-7);
+                return true;
+            case Month:
+                since = now.AddDays(-30);
+                return true;
+            case All:
+                since = DateTimeOffset.MinValue;
+                return true;
+            default:
+                since = default;
+                return false;
+        }
+    }
+}
--- a/server/src/GankedTV.Api/Contracts/Leaderboards/LeaderboardWindow.cs
+++ b/server/src/GankedTV.Api/Contracts/Leaderboards/LeaderboardWindow.cs
@@ -12,7 +12,10 @@ public static class LeaderboardWindow
     public const string Default = Week;
 
     // `since` for `all` is DateTimeOffset.MinValue so the same `l.CreatedAt >= since`
-    // filter covers every window with no branching at the query layer.
+    // filter covers every window with no branching at the query layer. Note the inclusive
+    // `>=`: trending uses strict `>` (see TryParseTrendingWindow) — inconsequential at
+    // second precision, but called out so a future reader copy-pasting between the two
+    // endpoints doesn't assume the boundary semantics are interchangeable.
     public static bool TryParse(string? window, out DateTimeOffset since)
     {
         var now = DateTimeOffset.UtcNow;
@@ -22,6 +25,10 @@ public static class LeaderboardWindow
                 since = now.AddDays(-7);
                 return true;
             case Month:
+                // Deliberately a rolling 30 days, not a calendar month — the web label
+                // ("This Month") is a simplification. A clip liked 31 days ago drops out
+                // of the window even mid-calendar-month. Renaming the key to `30d` would
+                // be more honest, but `month` matches the user-facing label.
                 since = now.AddDays(-30);
                 return true;
             case All:
--- a/server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs
+++ b/server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs
@@ -136,9 +136,11 @@ public static class GamesEndpoints
 
         // Resolve game first so "no such game" stays a 404 even when there are no likes in
         // the window — otherwise an unknown slug would silently return an empty board.
+        // Inline the GameSummary columns so EF projects rather than materializing the full
+        // entity (extension-method calls inside Select can't be translated).
         var game = await db.Games.AsNoTracking()
             .Where(g => g.Slug == slug)
-            .Select(g => new { g.Id, Summary = g.ToGameSummary() })
+            .Select(g => new { g.Id, Summary = new GameSummary(g.Id, g.Name, g.Slug, g.Tag) })
             .FirstOrDefaultAsync(ct);
 
         if (game is null)
--- a/server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs
+++ b/server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs
@@ -1,5 +1,6 @@
 using System.Security.Claims;
 using GankedTV.Api.Contracts.Games;
+using GankedTV.Api.Contracts.Leaderboards;
 using GankedTV.Api.Data;
 using GankedTV.Api.Problems;
 using GankedTV.Api.Services.ObjectStorage;
@@ -19,6 +20,7 @@ public static class GamesEndpoints
         group.MapGet("/", GetGames);
         group.MapGet("/{slug}", GetBySlug);
         group.MapGet("/{slug}/clips", GetClipsForGame);
+        group.MapGet("/{slug}/leaderboard", GetLeaderboardForGame);
         return app;
     }
 
@@ -114,5 +116,41 @@ public static class GamesEndpoints
         var response = await ClipsReadEndpoints.BuildFeedPageAsync(
             baseQuery, cursor, limit, principal, db, storage, s3, ct);
         return Results.Ok(response);
+    }
+
+    private static async Task<IResult> GetLeaderboardForGame(
+        string slug,
+        string? window,
+        int? limit,
+        ClaimsPrincipal principal,
+        GankedTvDbContext db,
+        IObjectStorageService storage,
+        IOptions<S3Options> s3,
+        CancellationToken ct)
+    {
+        var windowKey = window ?? LeaderboardWindow.Default;
+        if (!LeaderboardWindow.TryParse(windowKey, out var since))
+        {
+            return ProblemResults.BadRequest("invalid_window");
+        }
+
+        // Resolve game first so "no such game" stays a 404 even when there are no likes in
+        // the window — otherwise an unknown slug would silently return an empty board.
+        var game = await db.Games.AsNoTracking()
+            .Where(g => g.Slug == slug)
+            .Select(g => new { g.Id, Summary = g.ToGameSummary() })
+            .FirstOrDefaultAsync(ct);
+
+        if (game is null)
+            return ProblemResults.NotFound("not_found");
+
+        var cap = Math.Clamp(limit ?? LeaderboardsEndpoints.DefaultClipsLimit, 1, LeaderboardsEndpoints.MaxClipsLimit);
+        var baseQuery = db.Clips.AsNoTracking()
+            .Where(c => c.GameId == game.Id && c.Visibility == "public" && c.Status == "ready");
+
+        var entries = await LeaderboardsEndpoints.BuildEntriesAsync(
+            baseQuery, since, cap, principal, db, storage, s3, ct);
+
+        return Results.Ok(new GameLeaderboardResponse(windowKey, game.Summary, entries));
     }
 }
--- a/server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
+++ b/server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
@@ -99,8 +99,13 @@ public static class LeaderboardsEndpoints
             .ToList();
 
         var topIds = ranked.Select(r => r.ClipId).ToList();
+        // Re-apply the public+ready predicate as a belt-and-braces guard: a clip can flip
+        // to private/unlisted or out of `ready` between the candidate fetch above and this
+        // hydrate. Trending tolerates that race ("self-healing on the next request"), but
+        // a leaderboard surfacing an unlisted clip in a "Most Liked This Week" board is
+        // more user-visible/confusing, and the extra WHERE clause is free.
         var hydrated = await db.Clips.AsNoTracking()
-            .Where(c => topIds.Contains(c.Id))
+            .Where(c => topIds.Contains(c.Id) && c.Visibility == "public" && c.Status == "ready")
             .IncludeFeedRelations()
             .ToListAsync(ct);
         var byId = hydrated.ToDictionary(c => c.Id);
@@ -111,8 +116,8 @@ public static class LeaderboardsEndpoints
 
         // Walk ranked + feedItems in lockstep: feedItems was built from ranked-ordered
         // clips, so index i corresponds to the same clip in both lists. A clip dropped
-        // between candidate fetch and hydrate (rare but possible mid-delete) just falls
-        // out of both lists together.
+        // between candidate fetch and hydrate (mid-delete, or filtered out by the
+        // re-applied visibility/status guard above) just falls out of both lists together.
         var entries = new List<LeaderboardEntry>(feedItems.Count);
         var rank = 0;
         var feedIndex = 0;
--- a/server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
+++ b/server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
@@ -1,0 +1,167 @@
+using System.Security.Claims;
+using GankedTV.Api.Contracts.Clips;
+using GankedTV.Api.Contracts.Games;
+using GankedTV.Api.Contracts.Leaderboards;
+using GankedTV.Api.Data;
+using GankedTV.Api.Data.Entities;
+using GankedTV.Api.Problems;
+using GankedTV.Api.Services.ObjectStorage;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Options;
+
+namespace GankedTV.Api.Endpoints;
+
+public static class LeaderboardsEndpoints
+{
+    internal const int DefaultClipsLimit = 10;
+    internal const int MaxClipsLimit = 50;
+    internal const int DefaultGamesLimit = 10;
+    internal const int MaxGamesLimit = 25;
+
+    public static IEndpointRouteBuilder MapLeaderboardsEndpoints(this IEndpointRouteBuilder app)
+    {
+        app.MapGet("/leaderboards", GetGlobal);
+        return app;
+    }
+
+    private static async Task<IResult> GetGlobal(
+        string? window,
+        int? clipsLimit,
+        int? gamesLimit,
+        ClaimsPrincipal principal,
+        GankedTvDbContext db,
+        IObjectStorageService storage,
+        IOptions<S3Options> s3,
+        CancellationToken ct)
+    {
+        // Default to "week" when the caller omits window — the standalone /leaderboards
+        // landing page renders week-first; explicit-but-unknown values still 400 so a UI
+        // typo (`?window=mounth`) surfaces immediately instead of being silently coerced.
+        var windowKey = window ?? LeaderboardWindow.Default;
+        if (!LeaderboardWindow.TryParse(windowKey, out var since))
+        {
+            return ProblemResults.BadRequest("invalid_window");
+        }
+
+        var clipsCap = Math.Clamp(clipsLimit ?? DefaultClipsLimit, 1, MaxClipsLimit);
+        var gamesCap = Math.Clamp(gamesLimit ?? DefaultGamesLimit, 1, MaxGamesLimit);
+
+        var clipsBase = db.Clips.AsNoTracking()
+            .Where(c => c.Visibility == "public" && c.Status == "ready");
+        var topClips = await BuildEntriesAsync(clipsBase, since, clipsCap, principal, db, storage, s3, ct);
+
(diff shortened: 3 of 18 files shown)


server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
```diff
@@ -0,0 +1,172 @@
...
+                WindowLikes = db.Likes.Count(l =>
+                    l.CreatedAt >= since
+                    && l.Clip.GameId == g.Id
+                    && l.Clip.Visibility == "public"
+                    && l.Clip.Status == "ready"),
+                ClipCount = db.Clips.Count(c =>
+                    c.GameId == g.Id && c.Visibility == "public" && c.Status == "ready"),
+            })
+            .Where(x => x.WindowLikes > 0)
+            .ToListAsync(ct);
```
Turbootzz, line 156: **Top-games materializes every active game.** The full `Game` entity is loaded for every game with `WindowLikes > 0`, then ordering+`Take(N)` happens in memory. For the `all` window that's "every game with ≥1 like, ever." Fine today; if the catalog grows past a few thousand active games this becomes chunky. Cheap mitigation: do `.OrderByDescending(...).ThenByDescending(...).Take(limit)` in SQL (the `Name Ordinal` tiebreak translates), and project just the columns needed for `TopGameEntry` instead of pulling the whole entity (skips `CoverUrl` when it isn't rendered — see related comment in `LeaderboardsView.vue`).

server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs
```diff
@@ -115,4 +117,40 @@ private static async Task<IResult> GetClipsForGame(
...
+        if (!LeaderboardWindow.TryParse(windowKey, out var since))
+        {
+            return ProblemResults.BadRequest("invalid_window");
+        }
+
+        // Resolve game first so "no such game" stays a 404 even when there are no likes in
+        // the window — otherwise an unknown slug would silently return an empty board.
+        var game = await db.Games.AsNoTracking()
+            .Where(g => g.Slug == slug)
+            .Select(g => new { g.Id, Summary = g.ToGameSummary() })
```
Turbootzz: **EF can't translate the extension method call here**, so it falls back to materializing the full `Game` entity and calling `ToGameSummary()` client-side. Tests pass because the materialization is implicit, but the code reads like a column-projection. Either inline `new GameSummary(g.Id, g.Name, g.Slug, g.Tag)`, or select the entity and map after `FirstOrDefaultAsync` — matches the existing pattern in `QueryGameDetailAsync` a few methods up.

server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
```diff
@@ -0,0 +1,172 @@
...
+public static class LeaderboardsEndpoints
+{
+    internal const int DefaultClipsLimit = 10;
+    internal const int MaxClipsLimit = 50;
+    internal const int DefaultGamesLimit = 10;
+    internal const int MaxGamesLimit = 25;
+
+    public static IEndpointRouteBuilder MapLeaderboardsEndpoints(this IEndpointRouteBuilder app)
+    {
+        app.MapGet("/leaderboards", GetGlobal);
```
Turbootzz, line 23: **No caching layer**, even though `/games/{slug}/clips` and `/games/{slug}` both go through `IFeedCache` / `IGamesCache`. Leaderboards are arguably *more* cacheable than feeds — results only change when likes happen and the ranking is approximate by design. Not blocking, but the embedded `GameLeaderboardBlock` will fire one extra cold query per game-page render. Worth a follow-up issue.

server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
```diff
@@ -0,0 +1,172 @@
...
+        var entries = new List<LeaderboardEntry>(feedItems.Count);
+        var rank = 0;
+        var feedIndex = 0;
+        foreach (var r in ranked)
+        {
+            if (!byId.ContainsKey(r.ClipId)) continue;
+            rank++;
+            entries.Add(feedItems[feedIndex].ToEntry(rank, r.WindowLikes));
+            feedIndex++;
+        }
```
Turbootzz, line 128: **Style nit / DRY:** `rank` and `feedIndex` always move together — every increment is gated on the same `if`. They're equal to `entries.Count` at every step. Could collapse to:

```csharp
foreach (var r in ranked)
{
    if (!byId.ContainsKey(r.ClipId)) continue;
    entries.Add(feedItems[entries.Count].ToEntry(entries.Count + 1, r.WindowLikes));
}
```

server/src/GankedTV.Api/Contracts/Leaderboards/LeaderboardWindow.cs
```diff
@@ -0,0 +1,35 @@
...
+    public static bool TryParse(string? window, out DateTimeOffset since)
+    {
+        var now = DateTimeOffset.UtcNow;
+        switch (window)
+        {
+            case Week:
+                since = now.AddDays(-7);
+                return true;
+            case Month:
+                since = now.AddDays(-30);
```
Turbootzz, line 32: **"Month" is 30 days, not a calendar month.** Standard simplification, but the label "This Month" the web renders implies otherwise (a clip liked 31 calendar-days ago would silently drop out mid-month). Worth either a one-liner comment noting the deliberate choice, or renaming the key to `30d` to match the trending vocabulary's day-based naming.

Turbootzz (commented):
Solid PR — correctness is well-tested and the trending-parity comments make intent clear. Inline notes below cover three buckets: a couple of structural items (#1 top-games materialization, #2 EF projection over-fetch), a handful of frontend nits, and three DRY opportunities now that the second windowed-ranking endpoint exists. None are blockers; the frontend nits (#4 unused CoverUrl, #5 dead prop, #6 comment/default mismatch) are the easiest to fold in before merge.

server/src/GankedTV.Api/Contracts/Leaderboards/LeaderboardWindow.cs
```diff
@@ -0,0 +1,35 @@
...
+// boards but never for trending. Kept separate from TryParseTrendingWindow so the two
+// endpoints can evolve their window vocabularies independently.
+public static class LeaderboardWindow
+{
+    public const string Week = "week";
+    public const string Month = "month";
+    public const string All = "all";
+    public const string Default = Week;
+
+    // `since` for `all` is DateTimeOffset.MinValue so the same `l.CreatedAt >= since`
```
Turbootzz, line 14: **Minor semantic drift vs trending**: leaderboards use `>= since`, trending uses `> since`. Inconsequential at second precision but worth a one-liner explaining it's a deliberate choice (matters especially when `since == DateTimeOffset.MinValue` for the `all` window — strict `>` would still match everything, but the asymmetry could trip up a future reader copy-pasting between the two endpoints).

server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    // Shared helper: turn a pre-filtered Clip IQueryable into a ranked list of leaderboard
+    // entries scoped to a window. Owns the windowed-count query, deterministic tiebreak,
+    // hydration with feed includes, and the rank-numbering pass.
+    //
+    // Tie-breaking goes (likes desc, clip.created_at desc, clip.id asc): newer clips with
+    // equal likes outrank older ones (rewards momentum); equal createdAt falls back to a
+    // total ordering on Guid so the ranking is stable across requests.
+    internal static async Task<List<LeaderboardEntry>> BuildEntriesAsync(
```
Turbootzz, line 65: **DRY — shape duplication with trending.** The comment on line 76 already acknowledges this: `BuildEntriesAsync` and `ClipsReadEndpoints.BuildTrendingFeedAsync` now share the same four-step shape (filter candidates with engagement in window → order in-memory by score+tiebreak → rehydrate with feed Includes → project). Differences are just the scoring expression (rank-by-windowed-likes vs. trending-score) and the belt-and-braces re-filter on hydrate. Worth a follow-up to extract a generic `BuildRankedFeedAsync<TScore>(baseQuery, since, scoreSelector, tiebreak, ...)` so the next windowed ranker (top-creators? top-tags?) drops in without copying the whole shape again.

server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs
```diff
@@ -115,4 +117,40 @@ private static async Task<IResult> GetClipsForGame(
...
+        GankedTvDbContext db,
+        IObjectStorageService storage,
+        IOptions<S3Options> s3,
+        CancellationToken ct)
+    {
+        var windowKey = window ?? LeaderboardWindow.Default;
+        if (!LeaderboardWindow.TryParse(windowKey, out var since))
+        {
+            return ProblemResults.BadRequest("invalid_window");
+        }
```
Turbootzz, line 135: **DRY — window-parse + limit-clamp boilerplate is now in two endpoints.** Same 8 lines also live at the top of `LeaderboardsEndpoints.GetGlobal`. Worth a tiny helper on `LeaderboardWindow` (or a private static in `LeaderboardsEndpoints`) like `TryParseRequest(string? window, int? limit, int defaultLimit, int maxLimit, out string windowKey, out DateTimeOffset since, out int cap)` so adding a third leaderboard endpoint doesn't re-derive this.

web/src/components/LeaderboardRow.vue
```diff
@@ -0,0 +1,64 @@
...
+  // instead of the clip's all-time `likeCount`. The standalone /leaderboards page
+  // wants the windowed number — that's what makes the ranking meaningful — while
+  // a global "popular this week" embed can opt back into the all-time count.
+  showWindowLikes?: boolean
+}>()
+
+const isTopThree = computed(() => props.entry.rank <= 3)
+const displayedLikes = computed(() =>
+  props.showWindowLikes === false ? props.entry.clip.likeCount : props.entry.windowLikes,
+)
```
Turbootzz: **Dead-code prop.** `showWindowLikes` is defined + computed but never set by either caller — both `GameLeaderboardBlock.vue` and `LeaderboardsView.vue` use the default. Project convention is "don't design for hypothetical future requirements" (CLAUDE.md). Recommend removing the prop and just rendering `entry.windowLikes` directly until a caller actually needs the all-time count.

web/src/components/GameLeaderboardBlock.vue
```diff
@@ -0,0 +1,86 @@
...
+
+const props = withDefaults(
+  defineProps<{
+    slug: string
+    window?: LeaderboardWindow
+    // Mini view shows the top-3 only — fits a sidebar/embed nicely. The full /leaderboards
+    // page uses the row component directly with its own limit.
+    limit?: number
+  }>(),
+  { window: 'week', limit: 5 },
```
Turbootzz: **Comment/default mismatch.** Comment says "top-3 only" but the default is `5`, and the only caller (`GameView.vue:255`) also passes `:limit="5"`. Either fix the comment or change the default to 3.

web/src/views/LeaderboardsView.vue
```diff
@@ -0,0 +1,173 @@
...
+
+            <div class="min-w-0 flex-1 flex flex-col gap-px">
+              <span class="font-body text-[13px] font-medium text-text-primary truncate">
+                {{ entry.game.name }}
+              </span>
+              <span class="font-mono text-[10px] text-text-muted">
+                {{ formatNum(entry.windowLikes) }} ♥ · {{ formatNum(entry.clipCount) }} clips
+              </span>
+            </div>
+          </RouterLink>
```
Turbootzz, line 177: **`TopGameEntry.coverUrl` is fetched but never rendered here.** The contract carries it (and `BuildTopGamesAsync` reads it from the DB), but the top-games rows are text-only. Either render the cover thumbnail next to the rank — it'd give the right column a lot more visual weight — or drop `CoverUrl` from `TopGameEntry` so the contract matches the rendering.

web/src/views/LeaderboardsView.vue
```diff
@@ -0,0 +1,173 @@
...
+    if (id !== latestLoadId) return
+    data.value = resp
+  } catch (err) {
+    if (id !== latestLoadId) return
+    console.error('leaderboards: load failed', err)
+    errored.value = true
+  } finally {
+    if (id === latestLoadId) loading.value = false
+  }
+}
```
Turbootzz, line 49: **DRY — `latestLoadId` pattern is now in 3 places.** Identical structure (monotonic id → set loading → await → stale-check → set data/error → stale-check on finally) appears in `TrendingView.vue`, `GameLeaderboardBlock.vue`, and here. Worth a composable in `web/src/composables/` like `useLatestRequest<T>(loader: () => Promise<T>)` returning `{ data, loading, errored, run }`. Would cut ~20 lines per call site and make the staleness invariant uniform.

web/src/components/AppNav.vue
```diff
@@ -287,6 +287,13 @@ onBeforeUnmount(() => {
         >
           Trending
         </RouterLink>
+        <RouterLink
+          to="/leaderboards"
+          class="relative rounded-sm px-3.5 py-2 text-[13px] font-medium uppercase tracking-[0.04em] text-text-secondary no-underline transition-colors duration-150 hover:bg-surface-overlay hover:text-text-primary max-tablet:hidden"
```
Turbootzz, line 292: **Mobile users can't reach `/leaderboards`.** `max-tablet:hidden` matches the Trending link's pattern and there's no mobile drawer that picks them up. Pre-existing gap rather than a regression — flagging it here because two top-level features are now mobile-invisible. Might be worth a small `MobileNavMenu` follow-up before a third one lands.

server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs
mintycw: Fixed in 96ff1f2 — inlined the `GameSummary` constructor so EF projects columns instead of materializing the full entity.

server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
mintycw: Fixed in 96ff1f2 — `entries.Count` now serves as both the next rank and the next feedItems index.

server/src/GankedTV.Api/Contracts/Leaderboards/LeaderboardWindow.cs
mintycw: Added a comment in 96ff1f2 noting the deliberate rolling-30-days choice. Renaming the key to `30d` would be more honest but the user-facing label is "This Month", so the asymmetry stays.

server/src/GankedTV.Api/Contracts/Leaderboards/LeaderboardWindow.cs
mintycw: Documented the `>= since` vs trending `> since` asymmetry in 96ff1f2.

web/src/components/LeaderboardRow.vue
mintycw: Removed in 96ff1f2 — CLAUDE.md says no speculative API surface, and you are right that no caller sets it.

web/src/components/GameLeaderboardBlock.vue
mintycw: Fixed the comment in 96ff1f2 — the default is 5 and matches the only caller, so the comment was the wrong half of the pair.

web/src/views/LeaderboardsView.vue
mintycw: Went with rendering — covers are already on the wire, and the right column did look text-heavy. Added a 3:4 thumbnail in 96ff1f2.

server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
mintycw: Agree this is worth doing. Spinning into a follow-up issue rather than this PR — moving the `Take(limit)` into SQL means dropping `StringComparer.Ordinal` on the `Name` tiebreak (or living with Postgres collation), so it has its own design decision attached.

server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
mintycw: Filing as a follow-up. Caching `/games/{slug}/leaderboard` is especially attractive since GameView always cold-fires it.

server/src/GankedTV.Api/Endpoints/LeaderboardsEndpoints.cs
mintycw: Agreed — `BuildRankedFeedAsync<TScore>` is a clean extraction. Adding to the follow-up issue so trending and leaderboards can share it.

server/src/GankedTV.Api/Endpoints/GamesEndpoints.cs
mintycw: Agreed — `LeaderboardWindow.TryParseRequest(...)` makes sense as a helper. Bundled into the follow-up issue.

web/src/views/LeaderboardsView.vue
mintycw: Agreed — `useLatestRequest<T>` fits cleanly under `web/src/composables/`. Filing as a follow-up so it lands in one sweep across trending + both leaderboard call sites.

web/src/components/AppNav.vue
mintycw: Agreed — two top-level features now mobile-invisible is the right time to do a mobile drawer. Pre-existing, so leaving this PR untouched and folding it into the follow-up issue.

mintycw:
Pushed 96ff1f2 addressing 7 of the inline threads (GameSummary projection, rank/feedIndex collapse, window comments, dead prop, comment/default fix, cover thumbs). Six items deferred to #117 (caching, top-games SQL ordering, ranked-feed extraction, window-parse helper, useLatestRequest composable, mobile nav drawer) — each replied inline.

Turbootzz (approved):

PR #116, opened 2026-05-24
[server][web] Clip of the Day: featured hero clip

### LINKED ISSUE #103: [server][web] Clip of the Day: featured hero clip
## Summary

Phase 4 (Scale & Grow) — surface a daily "Clip of the Day" in the home hero. Today [HomeView.vue](web/src/views/HomeView.vue)'s hero card just renders `items[0]` (the newest clip from the feed) — there's no editorial/algorithmic featured pick.

Per [PLAN.md §9](PLAN.md) Phase 4: *"Clip of the Day" featured clip*.

**Depends on #87** — reuses its time-weighted trending score to pick the winner.

## Server

- Daily featured selection reusing #87's time-weighted score over a 24h window (likes + views). Two acceptable approaches (implementer's call, document in PR):
  - **A — pinned per day:** an `IHostedService` (same pattern as the other hosted services in `Program.cs`) computes the winner once per day and stores it under a cache/Redis key (`featured:{yyyy-mm-dd}`). Stable for the whole day.
  - **B — computed endpoint:** `GET /clips/featured` computes top-by-score on read with a short cache.
- Excludes non-`public` / non-`ready` clips. Deterministic tie-break (e.g. higher `likeCount`, then newer `created_at`, then `id`).
- Endpoint: `GET /clips/featured` → a single hydrated clip (reuse [ClipMappings.cs](server/src/GankedTV.Api/Contracts/Clips/ClipMappings.cs)), or `204`/empty when the DB has no eligible clips.

## Web

- Add `clips.featured()` to [web/src/api/clips.ts](web/src/api/clips.ts).
- Wire the [HomeView.vue](web/src/views/HomeView.vue) hero card to the featured clip instead of `items[0]`; add a "Clip of the Day" label/badge. Fall back to newest clip if `featured` is empty so the hero never goes blank.

## How to test manually

```bash
curl -s http://localhost:5050/clips/featured | jq '{id, title, likeCount, viewCount}'
# call again → same clip within the same day
# Browser: / shows the hero badged "Clip of the Day"
```

## Checklist

- [ ] Featured pick is stable within a calendar day and rolls over the next day
- [ ] Only `public` + `ready` clips are eligible; deterministic tie-break
- [ ] Empty DB / no eligible clips renders the hero without crashing (newest-clip fallback)

Milestone: Phase 4 — Scale & Grow · Source: [PLAN.md §9](PLAN.md) · Depends on: #87

mintycw:
## Summary

Adds a daily "Clip of the Day" pick to the home page hero. Selection reuses the time-weighted trending score from #87 over a UTC-calendar-day window, restricted to public + ready clips, with a deterministic tie-break so a single winner is pinned for the whole day. Empty result returns 204 and the web falls back to the newest clip so the hero never goes blank.

## What's here

**Server (`GET /clips/featured`)** — `ClipsReadEndpoints.cs`
- Score: `(likes_today * 3 + views_today) / pow(hours_since_created + 2, 1.5)` — same formula as trending, just scoped to engagement since 00:00 UTC.
- Tie-break: `Score → LikeCount → CreatedAt → Id`, all DESC. Single deterministic winner per day.
- `IMemoryCache` keyed by `featured:{yyyy-MM-dd}` with `AbsoluteExpiration` at next UTC midnight. Caches only the winning `Guid`; hydration runs every request so `likedByMe` and presigned URLs stay per-caller / fresh.
- Stale-cache eviction: if the cached pick is unpublished/deleted between selection and rehydration, the key is dropped and that one request returns 204 (next request recomputes).
- 11 new integration tests covering eligibility filters, scoring, all three tie-break levels, cache pinning, stale eviction, and per-caller `likedByMe`.

**Web** — `clips.featured()` + HomeView
- `clips.featured()` returns `Promise<ClipFeedItem | null>` (normalizes the existing `client.ts` 204 → `undefined` behavior to an explicit `null`).
- HomeView loads featured in parallel with the feed via `Promise.allSettled` so a featured failure can't block the feed.
- Hero falls back to newest clip when featured is null. Badge label switches between "Clip of the Day" (real pick) and "Featured Clip" (fallback) so the UI doesn't lie about provenance.
- 2 new client tests.

**Docs**
- `docs/superpowers/specs/2026-05-24-clip-of-the-day-design.md` — design contract.
- `docs/superpowers/plans/2026-05-24-clip-of-the-day.md` — TDD implementation plan (used to drive the build).

Closes #103

## How to test manually

```bash
make up && make seed
make server &      # API on :5050 (or worktree port — `make ports`)
cd web && bun dev  # Web on :5173

# 1. Empty engagement → 204, hero falls back to newest with "Featured Clip" badge
curl -sw '\nHTTP %{http_code}\n' http://localhost:5050/clips/featured  # → 204

# 2. Generate engagement → 200, hero shows that clip with "CLIP OF THE DAY" badge
curl -X POST http://localhost:5050/clips/00000000-0000-0000-0000-000000000001/view
curl -s http://localhost:5050/clips/featured | jq '{id, title, viewCount}'

# 3. Cache pinning — second call returns the same id even if another clip pulls ahead
for i in 06 07 08 09 10; do
  for j in 1 2 3; do
    curl -s -X POST http://localhost:5050/clips/00000000-0000-0000-0000-0000000000$i/view -o /dev/null
  done
done
curl -s http://localhost:5050/clips/featured | jq '.id'  # still clip 01
```

Browser: load `/`, confirm the badge above the hero title and that switching Latest ↔ Following tabs doesn't blank the hero.

## Checklist

- [x] Server: 720/720 tests pass, coverage 95.68% line / 87.5% branch (gate 85/85)
- [x] Web: 190/190 tests pass, coverage 94.17% line / 95.06% branch (gate 85/85)
- [x] Verified manually in browser (cached pick, fallback path, badge swap)
- [x] `make ci` green

### DIFF
--- a/server/tests/GankedTV.Api.Tests/Integration/Endpoints/ClipsReadEndpointsTests.cs
+++ b/server/tests/GankedTV.Api.Tests/Integration/Endpoints/ClipsReadEndpointsTests.cs
@@ -6,6 +6,7 @@ using GankedTV.Api.Clips;
 using GankedTV.Api.Data.Entities;
 using GankedTV.Api.Services.ObjectStorage;
 using GankedTV.Api.Tests.TestSupport;
+using Microsoft.EntityFrameworkCore;
 using NSubstitute;
 
 namespace GankedTV.Api.Tests.Integration.Endpoints;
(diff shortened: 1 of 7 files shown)


server/tests/GankedTV.Api.Tests/Integration/Endpoints/ClipsReadEndpointsTests.cs
```diff
@@ -1162,4 +1163,361 @@ public async Task ShareCodeResolve_CrawlerUA_HtmlEscapesTitle()
...
+
+    [Fact]
+    public async Task Featured_PicksHighestScoringClip()
+    {
+        // Three clips of identical age — engagement alone decides. The clip with the
+        // most views in today's UTC window wins.
+        await _fx.ResetAsync();
+        var (userId, _) = await SeedUserAndIssueTokenAsync();
+        var now = DateTimeOffset.UtcNow;
+        var todayStart = now.Date;
```
Turbootzz: `now.Date` returns a `DateTime` with `Kind=Unspecified`. The comparison `now.AddMinutes(-5) >= todayStart` then implicitly converts that `DateTime` to `DateTimeOffset` using **local time**, not UTC — and so does the ternary result that becomes `engagementAt`.

In CI (UTC host) this works. On a non-UTC host (e.g. UTC+5 dev box) near midnight UTC, the fallback branch produces `engagementAt = 2026-05-24T00:00:00+05:00` = `2026-05-23T19:00:00Z`, which is *before* the production code's `todayStart` (true UTC midnight), so the seeded engagement would fall outside today's window and the test would fail.

The production code does this correctly with `new DateTimeOffset(DateTimeOffset.UtcNow.Date, TimeSpan.Zero)`. The tests should mirror it — this pattern is repeated in all of the new featured tests (~9 occurrences).

mintycw: Verified — `DateTimeOffset.UtcNow.Date` returns `DateTime` with `Kind=Unspecified`, and the implicit conversion in the ternary uses the local offset, so on a UTC+ host near UTC midnight `engagementAt` would land outside production's UTC-day window. I already had the correct pattern in `Featured_NoEngagementToday_Returns204` (line 1293) but missed propagating it.

Fixed all 9 sites to mirror production: `new DateTimeOffset(now.UtcDateTime.Date, TimeSpan.Zero)`. `engagementAt` is now consistently `DateTimeOffset` from both branches. 11/11 featured tests still pass. Pushed as e4ea2f5.

Turbootzz (approved):

PR #120, opened 2026-05-24
[server][web] Moderation: roles, content reporting, admin dashboard (#105)

### LINKED ISSUE #105: [server][web] Moderation: user roles + content reporting + admin dashboard
## Summary

Phase 4 (Scale & Grow) — user roles, content reporting, and an admin moderation dashboard. Today there is **no role concept**: [JwtService.cs](server/src/GankedTV.Api/Auth/Jwt/JwtService.cs) issues `Sub`/`name`/`Jti`/`email` claims only, there's no `User.Role` column, and every protected endpoint uses a flat `RequireAuthorization()`. Users also have no way to report abusive content.

Per [PLAN.md §9](PLAN.md) Phase 4: *Admin dashboard + moderation tools* and the abuse-prevention half of *Rate limiting & abuse prevention* (rate limiting itself shipped in #74/#82; its Redis/cluster migration is tracked separately in the Redis issue).

## Server

- **Roles:** add `User.Role` (`user` | `moderator` | `admin`, default `user`) to [User.cs](server/src/GankedTV.Api/Data/Entities/User.cs) + migration. Emit a `role` claim from [JwtService.cs](server/src/GankedTV.Api/Auth/Jwt/JwtService.cs). Register `"admin"` / `"moderator"` authorization policies (config/DI in `Program.cs`).
- **Reports:** new `reports` table + `Report` entity (`reporter_id`, target — one of `clip_id` / `comment_id` / `user_id`, `reason`, `status` = `open|resolved|dismissed`, timestamps). `POST /clips/{id}/report` (+ comment/user variants), authenticated, rate-limited via the existing clips limiter ([ClipsRateLimiting.cs](server/src/GankedTV.Api/Clips/ClipsRateLimiting.cs)). **Self-report guard.**
- **Admin/mod endpoints** (`RequireAuthorization("moderator")` / `"admin"`):
  - `GET /admin/reports?status=open` — paginated queue with hydrated target + reporter.
  - `POST /admin/reports/{id}/resolve` — set status (`resolved`/`dismissed`).
  - Moderation actions: hide/remove a clip (set `visibility`/`status` so it drops from feeds), ban a user.
- Seed the first admin via env (e.g. `ADMIN_EMAILS`) or the dev seed script.
- Expose `role` on the `/me` response so the client can gate UI.

## Web

- **Report button** on [ClipView.vue](web/src/views/ClipView.vue) (and comments) → reason modal → `POST .../report`.
- **`/admin` route** + `AdminView.vue` gated by a router guard reading the `role` from the auth store ([web/src/stores/auth.ts](web/src/stores/auth.ts) / `MeResponse`). Non-admins are redirected.
- Admin nav link in [AppNav.vue](web/src/components/AppNav.vue), visible only to admins/mods.
- Reports queue UI: list, open target, resolve/dismiss, take a moderation action. New `web/src/api/admin.ts`.

## How to test manually

```bash
# Report (as a normal user)
curl -X POST -H "Authorization: Bearer $TOKEN" \
  http://localhost:5050/clips/<id>/report -d '{"reason":"spam"}'

# Admin queue (as admin)
curl -H "Authorization: Bearer $ADMIN_TOKEN" 'http://localhost:5050/admin/reports?status=open' | jq
# As a normal user → 403
curl -H "Authorization: Bearer $TOKEN" http://localhost:5050/admin/reports -i   # 403

# Hide a clip → confirm it disappears from GET /clips/feed
```

## Checklist

- [ ] Non-admin/mod gets 403 on `/admin/*`; admin nav link hidden for them
- [ ] Reporting creates a row; self-report is rejected
- [ ] Resolving/dismissing updates report status
- [ ] Hiding/removing a clip drops it from all feeds
- [ ] `role` claim present in JWT and surfaced on `/me`

Turbootzz:
Closes #105.

## Summary

Phase 4 moderation surface — user roles, content reporting, admin dashboard.

**Server**
- `User.Role` (`user` / `moderator` / `admin`) + `BannedAt` / `BannedReason`. `role` claim on every JWT, `MapInboundClaims=false` so the short name lands as-is. New `RoleAuthorization.AddRolePolicies()` registers `"admin"` and `"moderator"` (admin satisfies both).
- Polymorphic `Report` entity — one table for clips / comments / users with a partial unique index `(reporter_id, target_type, target_id) WHERE status='open'` for race-safe dedup.
- `POST /{clips,comments,users}/{id}/report` (auth + clip rate limiter) — all three share `ReportService.CreateAsync` with self-report + duplicate + missing-target guards.
- `/admin/*` under the moderator-tier group: queue with hydrated reporter+target, resolve/dismiss, hide/unhide clip, remove comment, **fix game tag** (reason-scoped auto-resolve closes only `wrong_game` reports), ban/unban (admin-only).
- `BannedUserMiddleware` rejects existing JWTs on the next request (401 + `account_banned`); login/refresh/OAuth callback all refuse to mint tokens for banned accounts.
- `AdminBootstrap` hosted service promotes `ADMIN_EMAILS` to admin idempotently. `make seed` ships a pre-promoted `seedadmin`. `/dev/token` accepts an allow-listed `role` for the dev sign-in button.
- Owner profile filter now shows public + unlisted to the owner viewing their own profile; foreign viewers still see public only; hidden stays invisible to all.

**Web**
- Auth store gets `isAdmin` / `isModerator`; router guard reads `meta.requiresRole`.
- One `ReportDialog.vue` reused from `ClipView` / `CommentItem` / `UserView` (target type + id are props).
- `/admin` route + `AdminView.vue` with tab-deep-linked queue, `ReportRow.vue` rendering target-aware actions, `FixGameDialog.vue` reusing the existing `GameSelector`.
- `KebabMenu.vue` extracted — own-profile **Sign out** + clip-owner Edit/Delete now share the same dropdown component with a `variant: 'danger'` (red) item style.
- "Sign in as seedadmin" dev button.

**Reasons:** `spam` / `harassment` / `hate` / `nsfw` / `violence` / `wrong_game` / `other` (`other` requires a note).

## Test plan

- [x] `make ci` is green (849 server tests + coverage gate 85/85, 201 web tests + coverage gate 85/85).
- [x] `make seed`, then sign in as `seedadmin` — Admin link appears in nav; non-admin doesn't see it.
- [x] As `seeduser2`, report any clip/comment/user. Self-report your own → 400; duplicate report → 409.
- [x] As admin, `/admin/reports` queue lists hydrated targets. Resolve / dismiss flip status into the right tab.
- [x] Hide clip → drops from `/clips/feed`; report auto-resolves. Unhide restores.
- [ ] Submit a `wrong_game` report → admin sees **Fix game** button → typeahead/popular-chips picker → save → clip's game_id updates; only wrong_game reports for that clip resolve.
- [ ] Ban a user → their existing JWT returns 401 `account_banned` on the next request; SPA flushes session.
- [ ] Owner sees their unlisted clips on their own profile; foreign viewer doesn't.
- [ ] Sign out from the profile-page kebab works.

## Known follow-ups (out of scope here)

- Hidden clips are still reachable by direct `/clip/{id}` / `/c/{code}` (visibility is enforced at listing layer only; pre-existing design choice). Adding `Visibility != "hidden"` to those three predicates is a small follow-up.

## Summary by CodeRabbit

* **New Features**
  * Report UI to submit reports on clips, comments, and users (reason + optional note).
  * Admin/moderator dashboard to list, filter, page, and act on reports (resolve/dismiss, hide/unhide clips, fix game tag, remove comments, ban/unban users).
  * Role-aware client: /auth/me returns role; admin link shown for moderators/admins; dev sign-in can create/return admin role.

* **Behavior changes**
  * Owners see public + unlisted clips; hidden clips are excluded for others.
  * Banned accounts are blocked from auth/refresh and from creating reports.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/120?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


mintycw (commented):
Review found 3 Important issues worth addressing before merge — none Critical, architecture is sound. Posting as a comment review rather than a request-for-changes so you can decide which to land in this PR vs. follow-up.

## Important (recommend fixing pre-merge)

### 1. `AdminBootstrap` never demotes when an email is removed from `ADMIN_EMAILS`
`server/src/GankedTV.Api/HostedServices/AdminBootstrap.cs:38-43`

The query promotes any email in the list to admin but takes no action on emails NOT in the list — so removing a user from `ADMIN_EMAILS` and restarting leaves them admin forever. The class-header comment notes the "promote on next login" behavior but is silent on demotion.

This is a governance/security gap: an operator who thinks they've revoked admin by editing the env var has not done so. Either:
- Document explicitly that demotion is manual (one-line update to the class comment), **or**
- Add a `promoted_by_bootstrap` flag and a second `ExecuteUpdateAsync` that demotes rows previously promoted by bootstrap whose email is no longer in the list (avoids demoting hand-promoted admins).

### 2. Ban button is rendered to moderators in `ReportRow.vue`
`web/src/components/admin/ReportRow.vue:118-125`

The Ban/Unban button shows for any `targetType === 'user'` report regardless of `auth.isAdmin`. Server admin-gates it correctly, so a moderator clicking it gets a 403 toast (`Action failed (403)`). The auth store already exposes `isAdmin` vs `isModerator` — `AppNav` and the router both gate on it.

Fix: thread `isAdmin` from `AdminView` into `ReportRow` (or import the auth store directly in `ReportRow`) and conditionally render the ban button. Same gating should be considered for any other admin-only-but-moderator-visible action in the row.

### 3. Banned user can still rotate refresh tokens (write-amplification)
`server/src/GankedTV.Api/Endpoints/AuthEndpoints.cs:251-261`

`refreshTokens.RotateAsync` runs **before** the ban check in `/auth/refresh`, so each call burns the old refresh AND writes a successor row, even though the endpoint then 403s. The inline comment acknowledges the successor is "unusable to a banned account", but it's still persisted.

A banned client polling `/auth/refresh` in a tight loop fills `refresh_tokens` indefinitely (one row per call). `/auth/refresh` has no rate-limit policy attached (unlike `/auth/login` → `AuthRateLimiting.CredentialsPolicy`), so this is driveable without per-IP throttling. The Maintenance sweeper GCs expired tokens but live rows for a banned user are pure write amplification.

Fix: check `BannedAt` **before** calling `RotateAsync` (in `AuthEndpoints.Refresh` or inside `RefreshTokenService.RotateAsync`) and 403 without issuing a new row.

## Minor / follow-ups (won't block)

- Reports share `ClipsWritePolicy` (30/min) with the entire clip-mutation surface. The PR notes this is intentional, but the rate-limit comment block in `Clips/ClipsRateLimiting.cs:21-31` enumerates which endpoints share the bucket and doesn't list reports — docs drift. Either give reports their own per-user bucket or extend the comment.
- No `banned_by` (Guid, nullable) audit column on `users` — can't tell which admin banned whom. Same for `unbanned_by` / `unhidden_by` on the corresponding endpoints.
- `BanUserRequest.Reason` is `[StringLength(500)]` but not `[Required]`; null/whitespace reasons defeat the audit trail.
- Hand-edited migration `20260524141621_AddModerationRolesReports` is correct but if someone re-runs `dotnet ef migrations add` they'll get a confusing diff — worth a top-of-file "do not regenerate" note.
- A `Task.WhenAll`-style concurrency test for `ResolveAsync`'s CAS path would pin the new guarantee (currently only the sequential "already resolved → 409" case is covered).
- A round-trip test that the DB-level `ck_reports_other_note` CHECK actually rejects a direct `Reports.Add(...)` with `reason="other"`, `note=null` would pin the defense-in-depth, since only the service-layer enforcement is currently tested.
- Stale comment in `AdminView.vue:111-114` claims `FixGameDialog` is "seeded with the clip's current game id" — it always starts empty.

## Strengths (for the record)

- Polymorphic Report with `(reporter_id, target_type, target_id) WHERE status='open'` partial unique index is the right race-safe dedup shape.
- CAS via `ExecuteUpdateAsync` is the right fix for concurrent moderators on resolve.
- `MapInboundClaims=false` wiring is consistent across issue / validate / policy paths; no `IsInRole` / `ClaimTypes.Role` usage to break.
- `BannedUserMiddleware` ordering is correct (between auth and authz), with login/refresh/OAuth callback all refusing to mint tokens for banned accounts.
- Tests assert real DB state (`Visibility == Hidden`, `BannedAt != null`, report `Status == Resolved`) rather than just status codes. The `SetClipGame_NullGameId_ClearsTag` tightening shows discipline.
- `KebabMenu` extraction unifies ClipView/UserView call sites cleanly via `iconOrientation` + `triggerVariant`; `ReportDialog` is genuinely reused across all three target types.
- `AdminView.load()` request-token pattern mirrors `UserView.latestLoadId` correctly.

**Verdict:** items 1, 2, 3 are the load-bearing ones. If you address those, the rest can land as follow-ups.

PR #121, opened 2026-05-25
[server][web] Leaderboards + trending follow-ups (#117)

### LINKED ISSUE #117: Leaderboards + trending: shared infra, caching, and mobile nav follow-ups
Follow-ups from PR #114 review (deferred so the initial leaderboards landing is small + reviewable).

## Server

- [ ] **Cache leaderboards.** `/games/{slug}/leaderboard` is cold-fired on every `GameView` render and is more cacheable than feeds (only changes when likes happen). Plumb through `IFeedCache`-shape cache; `/leaderboards` too.
- [ ] **Top-games: move ordering into SQL.** `BuildTopGamesAsync` currently materializes every game with ≥1 windowed like and orders + `Take(limit)` in memory. Fine today; chunky if the catalog grows past a few thousand active games. Requires deciding what to do with the `StringComparer.Ordinal` tiebreak (drop it or accept Postgres collation order).
- [ ] **Extract `BuildRankedFeedAsync<TScore>`.** `LeaderboardsEndpoints.BuildEntriesAsync` and `ClipsReadEndpoints.BuildTrendingFeedAsync` share the same 4-step shape (filter candidates with engagement in window → order in-memory by score+tiebreak → rehydrate with feed includes → project). Differences are the scoring expression and the leaderboards belt-and-braces re-filter on hydrate.
- [ ] **`LeaderboardWindow.TryParseRequest(window, limit, defaults, max, out ...)`.** The window-parse + limit-clamp boilerplate is now duplicated at the top of both `GetGlobal` and `GetLeaderboardForGame`.

## Web

- [ ] **`useLatestRequest<T>` composable.** The monotonic `latestLoadId` + stale-check pattern lives in three call sites now: `TrendingView`, `GameLeaderboardBlock`, `LeaderboardsView`. Worth a composable under `web/src/composables/` returning `{ data, loading, errored, run }`.
- [ ] **Mobile nav drawer.** Trending and Leaderboards top-level routes are both `max-tablet:hidden`, so mobile users can't reach them. Pre-existing gap; two top-level features mobile-invisible is the right time to build a `MobileNavMenu`.

Original review threads on PR #114 for context.

mintycw:
Closes #117 — the deferred refactors from PR #114.

## Summary

**Server**
- TTL-only `IFeedCache.GetOrCreateLeaderboardAsync<T>` wraps both `/leaderboards` and `/games/{slug}/leaderboard`. Entries are cached anonymously and re-stamped with `LikedByMe` per request (same pattern trending already uses) so the cache can't leak likes across users.
- `BuildEntriesAsync` split into `BuildAnonymousEntriesAsync` + `StampLikedByMeOnEntriesAsync`.
- New `RankedFeedBuilder.HydrateOrderedAsync` extracts the shared hydrate-and-preserve-order step from leaderboards + trending, with an optional `public+ready` re-filter (leaderboards opts in to guard the candidate→hydrate micro-race; trending opts out and self-heals via TTL).
- `BuildTopGamesAsync` now pushes `(WindowLikes desc, ClipCount desc)` into SQL with `Take(limit*4)`; the `StringComparer.Ordinal` name tiebreak runs in-memory on the bounded candidate set.
- `LeaderboardWindow.TryParseRequest` bundles window-parse + limit-clamp; `GetGlobal` and `GetLeaderboardForGame` adopt it.

**Web**
- `useLatestRequest<T>` composable replaces the open-coded `latestLoadId` stale-check in `TrendingView`, `LeaderboardsView`, and `GameLeaderboardBlock`.
- `MobileNavDrawer` + hamburger in `AppNav` (visible below the 720px tablet breakpoint) so Games / Trending / Leaderboards are reachable on mobile. Escape / backdrop / route-change close; focus returns to the hamburger.

## Decisions

- **Cache invalidation:** TTL-only (like trending), not invalidated on like writes. Likes are too high-frequency to bust the cache on every one; the 60s/30s window matches the staleness already accepted on cached feed cards.
- **Top-games tiebreak:** kept `StringComparer.Ordinal` via the hybrid SQL+in-memory sort, so existing tests stay green.
- **No `AbortController` in `useLatestRequest` v1:** current call sites don't cancel, and adding it would change semantics (in-flight requests would reject rather than silently discard).

## Test plan

- [x] `make ci-server` — 959 passed, 94.7% line / 85.3% branch (above the 85/85 gate).
- [x] `make ci-web` — 245 passed, 93.5% statements / 94.2% branch on the gated scope.
- [x] New unit tests: 8 for `useLatestRequest`, 2 for `IFeedCache.GetOrCreateLeaderboardAsync` (cache-hit + tag isolation).
- [x] New integration tests: `Global_Cached_LikedByMeStaysPerCaller` + `GameLeaderboard_Cached_LikedByMeStaysPerCaller` — anonymous request warms the cache, then an authenticated liker hits the same entry and still sees their own `likedByMe=true`.
- [x] Manual smoke at 390×800: hamburger appears, drawer opens with all four routes, Esc + backdrop + route-change all close it, focus returns to the hamburger.
- [x] Manual smoke at 1280×800: desktop nav unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added mobile navigation drawer for improved mobile experience

* **Performance Improvements**
  * Implemented leaderboard caching to reduce load times
  * Optimized trending feed hydration logic

* **Bug Fixes**
  * Ensured user-specific "liked by me" status remains accurate across cached leaderboards

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/121?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/src/GankedTV.Api/Contracts/Leaderboards/LeaderboardWindow.cs
+++ b/server/src/GankedTV.Api/Contracts/Leaderboards/LeaderboardWindow.cs
@@ -39,4 +39,24 @@ public static class LeaderboardWindow
                 return false;
         }
     }
+
+    // Bundles the window-default-and-parse + limit-clamp that every leaderboard handler
+    // does up-front. `windowKey` is the resolved string (defaulted when null) so callers
+    // can echo it back in the response payload; `since` is the parsed cutoff; `clampedLimit`
+    // is the input limit clamped to [1, maxLimit] with the default applied when null.
+    // Returns false only when an explicit, non-null window string doesn't match a known
+    // value — the caller should turn that into a 400.
+    public static bool TryParseRequest(
+        string? window,
+        int? limit,
+        int defaultLimit,
+        int maxLimit,
+        out string windowKey,
+        out DateTimeOffset since,
+        out int clampedLimit)
+    {
+        windowKey = window ?? Default;
+        clampedLimit = Math.Clamp(limit ?? defaultLimit, 1, maxLimit);
+        return TryParse(windowKey, out since);
+    }
 }
--- a/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
+++ b/server/src/GankedTV.Api/Endpoints/ClipsReadEndpoints.cs
@@ -10,6 +10,7 @@ using GankedTV.Api.Data.Entities;
 using GankedTV.Api.Pagination;
 using GankedTV.Api.Problems;
 using GankedTV.Api.Services.Caching;
+using GankedTV.Api.Services.Feeds;
 using GankedTV.Api.Services.Media;
 using GankedTV.Api.Validation;
 using GankedTV.Api.Services.ObjectStorage;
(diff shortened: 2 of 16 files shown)


mintycw:
Addressed the inline + nitpick comments in 54055b2:

**Applied**
- `GameLeaderboardBlock.vue` — added `props.limit` to the watch deps so limit changes refetch.
- `MobileNavDrawer.vue` — bumped wrapper from `z-50` → `z-[60]` so the drawer stacks explicitly above the `z-50` sticky header and matches the search dropdown / bell popover overlay layer.
- `LeaderboardsEndpointsTests.cs` — both `*_Cached_LikedByMeStaysPerCaller` tests now also warm the cache with an authenticated request first, then assert anonymous sees `likedByMe=false`, then re-assert the original direction. Covers cache-leak regressions in either direction.

**Skipped** (with reason)
- Outside-diff comment on `LeaderboardsEndpoints.cs` `BuildTopGamesAsync` (candidateCap → two-step cutoff-and-fetch): the `limit * 4` headroom is intentional and documented at line 169 as a trade-off. A boundary tie cluster of >30 games sharing identical `(WindowLikes, ClipCount)` at the cut is implausible enough that the added query round-trip + complexity isn't justified. Current behavior matches the in-code rationale.

Validated: `dotnet build server`, the two updated integration tests (both pass), and `bun run type-check`.

Turbootzz (approved):
Looks very cool, but before merging make sure its mobile responsive. We've got some work to do for responsiveness on some pages

Issue #123, opened 2026-05-26
[infra][ci] Containerize all services, publish to ghcr.io, and auto-deploy to host on merge to main


Turbootzz:
Phase 4 is nearly done. The repo has production-ready Dockerfiles for the API and Discord bot but no path to production: CI workflows only build/test, there's no web Dockerfile, no production compose, and nothing publishes images. This issue containerizes all services, publishes to **ghcr.io** on merge to `main`, and **auto-deploys to the host** via SSH.

**Blocked by #122** — `/health/ready` and startup migrations from that issue are required by the deploy job's rollout + smoke test.

External routing (domain + TLS) is handled by the user's existing **nginx proxy manager**, so containers here stay internal — no TLS in the images/compose.

## Tasks

- [ ] **`web/Dockerfile`** — multi-stage: `oven/bun` build (`bun install --frozen-lockfile` → `bun run build`) → **Caddy** static-serve stage (`caddy:alpine`) serving `dist/` with SPA fallback + gzip; **internal only, no TLS** (sits behind nginx proxy manager). Accept `VITE_API_BASE_URL` as a build arg. (`nginx:alpine` is an acceptable alternative — Caddy chosen for the simpler SPA config.)
- [ ] **Release/publish workflow** — `.github/workflows/release.yml`, trigger `push` to `main`. `permissions: { contents: read, packages: write }`. Use `docker/login-action` (registry `ghcr.io`, `GITHUB_TOKEN`) + `docker/metadata-action` + `docker/build-push-action` to build & push three images — `ghcr.io/gankedtv/gankedtv-api`, `-web`, `-discord` — each tagged with the commit SHA **and** `latest`. Build server from `server/Dockerfile.api`, discord from `discord/Dockerfile`, web from the new `web/Dockerfile`.
- [ ] **`docker-compose.prod.yml`** — references the ghcr.io images (not local build contexts): `api`, `web`, `discord` (profile-gated), plus `postgres`, `minio`, `redis`. Restart policies, named volumes, **no default/hardcoded credentials** (env-supplied), internal network only — external routing handled by nginx proxy manager (no TLS exposed here).
[...]

PR #125, opened 2026-05-26
[server][web] Production hardening: health checks, startup migrations, prod config, secret validation & GA4

### LINKED ISSUE #122: [server][web] Production hardening: health checks, startup migrations, prod config & dev-endpoint lockdown
Phase 4 is nearly complete and we're approaching the first deployment. Before any deploy can succeed safely, a handful of runtime gaps need closing. This is the **blocker** issue for the deployment work (see the CI/CD issue) — `/health/ready` and startup migrations in particular are required by the deploy job.

## Tasks

- [ ] **Health endpoints** — add ASP.NET Core health checks (`builder.Services.AddHealthChecks()` + a DbContext/Npgsql readiness check), map `GET /health/live` (process up, no deps) and `GET /health/ready` (DB reachable + migrations applied). Keep `Program.cs` wiring thin; put any real logic in a service so it stays inside the coverage denominator. Used by the deploy smoke-test and orchestration.
- [ ] **Startup DB migrations** — apply `dbContext.Database.Migrate()` on boot, gated behind an env flag (e.g. `RUN_MIGRATIONS_ON_STARTUP=true`, default off in dev) so a fresh prod DB self-migrates. Today migrations are manual (`make migrate`) — a fresh prod DB would let the API boot then fail on first query. `/health/ready` should only go green once migrations are applied.
- [ ] **`appsettings.Production.json`** — production defaults (logging levels, worker toggles per host role per the CLAUDE.md media-pipeline table). No secrets in the file.
- [ ] **Fail-fast secret validation** — when `ASPNETCORE_ENVIRONMENT=Production`, validate required secrets are present at startup (`DATABASE_URL`, `JWT_SECRET` ≥32B, `S3_*`, `WEB_ORIGIN`, `CORS_ORIGINS`) and refuse to boot with a clear error instead of running misconfigured.
- [ ] **Lock down dev endpoints** — ensure `POST /dev/token` is impossible outside Development (currently gated only by `IsDevelopment()` in `Program.cs:495`, so it's exposed if `ASPNETCORE_ENVIRONMENT` isn't explicitly `Production`). Deploy must set `ASPNETCORE_ENVIRONMENT=Production`; consider an extra explicit guard.
- [ ] **Web prod config** — confirm `VITE_API_BASE_URL` is injected at build time (build arg) so `dist/` points at the prod API, not `localhost:5050` (`web/src/api/client.ts`).
- [ ] **Deployment env docs** — `DEPLOYMENT.md` (or extend `.env.example`) listing all required prod env vars + how secrets are provided.

- [ ] **Web analytics (Google Analytics)** — wire up Google Analytics (GA4) in the web app, with the measurement ID injected via a `VITE_` env var (e.g. `VITE_GA_MEASUREMENT_ID`) so it's only active in production builds and easy to disable. Gate loading behind the presence of the ID so dev/local stays clean. Keep the integration behind a thin wrapper/composable so the analytics provider can be swapped later — we may expand to a product-analytics tool like PostHog down the line.

Turbootzz:
Closes #122 — the blocker for the CI/CD deploy work (#123).

## Server
- **Health endpoints** — `GET /health/live` (liveness, no deps) and `GET /health/ready` (DB reachable + migrations applied). Mapped as **terminal middleware before `UseHttpsRedirection`** so HTTP probes (container `HEALTHCHECK`, k8s, the #123 smoke-test) get `200`/`503` instead of a `307` redirect. Logic lives in `ReadinessHealthCheck`.
- **Startup migrations** — `DatabaseMigrator` runs `Database.MigrateAsync()` at boot, gated by `RUN_MIGRATIONS_ON_STARTUP` (default off; locally migrations stay manual). `/health/ready` only greens once migrations are applied.
- **`appsettings.Production.json`** — prod logging levels + main-API media-worker toggles. No secrets.
- **Fail-fast secret validation** — `ProductionStartupValidator` refuses to boot in Production when `DATABASE_URL`, `JWT_SECRET` (≥32B), `WEB_ORIGIN`, `CORS_ORIGINS`, or `S3_*` are missing/dev-default, with a clear aggregated error.
- **Dev-endpoint lockdown** — defense-in-depth runtime `IsDevelopment()` guard inside `POST /dev/token` (already unmapped by default outside dev).

## Web
- **GA4 analytics** behind a swappable wrapper (`src/lib/analytics.ts`), gated on `VITE_GA_MEASUREMENT_ID` — a complete no-op (no script, no cookies) when unset. SPA page views via `router.afterEach` (sends GA4 `page_location`/`page_title`). Env types added to `env.d.ts`.

## Docs
- New `DEPLOYMENT.md` (required prod env vars, secret handling, health endpoints, worker toggles, web build-time vars) + `.env.example` entries.

## Scope / follow-ups
- Containerization (`web/Dockerfile`, `docker-compose.prod.yml`, `release.yml`, deploy job) is **#123**, not here — the web build task here is docs + `.env.example` only.
- **Cookie-consent gating for GA is a documented follow-up** (GA currently loads whenever the measurement id is present).

## Verification
- Server: `make ci-server` green — 980 tests pass, coverage **94.71% line / 85.44% branch** (gate 85/85).
- Web: `make ci-web` green — type-check, lint, 244 tests, coverage gate met, prod build succeeds.

## Summary by CodeRabbit

* **New Features**
  * Fail-fast production startup validation for missing/invalid config.
  * Readiness and liveness health endpoints for monitoring.
  * Optional automatic DB migrations on startup (env-controlled).
  * Client analytics with GA4 support, safer page-view tracking that strips sensitive OAuth params.
  * Dev-only token endpoint now blocked outside development.

* **Documentation**
  * Added detailed production deployment guide and updated example env file.

* **Tests**
  * Added tests for startup validation, migrations, health checks, and analytics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/gankedtv/gankedtv/pull/125?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -1,0 +1,86 @@
+# Deployment
+
+Production runtime requirements and configuration for GankedTV. This covers the **API server**
+and the **web frontend**; container build/publish and the auto-deploy pipeline are tracked
+separately in issue #123.
+
+> **Secrets are never committed.** Everything below is provided via environment variables (or
+> your orchestrator's secret store) at runtime — there are no secrets in `appsettings*.json` or
+> any checked-in file.
+
+## Hard requirements
+
+- **`ASPNETCORE_ENVIRONMENT=Production` is mandatory.** It activates `appsettings.Production.json`,
+  the fail-fast secret validation, strict CORS, HTTPS redirection, and disables the dev-only
+  `POST /dev/token` and OpenAPI endpoints. A deploy that forgets this runs with dev affordances exposed.
+- **PostgreSQL ≥ 15** (the dev stack and CI run PG18).
+- The API image ([server/Dockerfile.api](server/Dockerfile.api)) ships `ffmpeg` + `yt-dlp` for the
+  media/import workers.
+
+## Fail-fast secret validation
+
+On boot in Production the API validates required configuration and **refuses to start** with an
+aggregated error if anything is missing or still set to a dev default (logic in
+[ProductionStartupValidator](server/src/GankedTV.Api/Configuration/ProductionStartupValidator.cs)).
+Required:
+
+| Var | Requirement |
+|---|---|
+| `DATABASE_URL` | Postgres connection string (dotnet/Npgsql form). |
+| `JWT_SECRET` | ≥ 32 bytes. Generate: `openssl rand -hex 32`. |
+| `WEB_ORIGIN` | Public origin of the web app (e.g. `https://ganked.tv`). |
+| `CORS_ORIGINS` | Comma-separated browser-origin allowlist. `WEB_ORIGIN` is always allowed implicitly. |
+| `S3_ENDPOINT` | Object-storage endpoint. |
+| `S3_ACCESS_KEY` / `S3_SECRET_KEY` | Non-default credentials (literal `minioadmin` is rejected). |
+| `S3_PUBLIC_URL` | Public base URL for stored objects / presigned-URL host rewrite. |
+
+Also recommended in production: `OAUTH_STATE_SECRET` (≥ 32 bytes; required once any OAuth provider
+is configured) and the provider client credentials (`DISCORD_*`, `GOOGLE_*`).
+
+## Startup database migrations
+
+Migrations are **manual locally** (`make migrate`). In production, set
+`RUN_MIGRATIONS_ON_STARTUP=true` on the migrating instance so a fresh DB self-migrates at boot,
+before serving requests. `/health/ready` only reports healthy once all migrations are applied
+(see [DatabaseMigrator](server/src/GankedTV.Api/Data/DatabaseMigrator.cs) and
+[ReadinessHealthCheck](server/src/GankedTV.Api/Services/Health/ReadinessHealthCheck.cs)). If you run
+more than one API replica, enable the flag on exactly one (or run migrations as a separate
+init step) to avoid concurrent migration races.
+
+## Health endpoints
+
+| Endpoint | Meaning | Use |
+|---|---|---|
+| `GET /health/live` | Process is up (no dependency checks). | Liveness probe / restart signal. |
+| `GET /health/ready` | DB reachable **and** migrations applied. | Readiness gate, load-balancer membership, the #123 deploy smoke test. |
+
+## Worker toggles (media pipeline)
+
+The media pipeline can be split across hosts (see the table in [CLAUDE.md](CLAUDE.md) under
+"Media pipeline"). `appsettings.Production.json` defaults to the **main API server** role
+(thumbnail worker on, transcode worker off, transcode pipeline on). Override per host via env vars
+to move GPU work to a separate box:
+
+| Instance | `MEDIA_THUMBNAIL_WORKER_ENABLED` | `MEDIA_TRANSCODE_WORKER_ENABLED` | encoder vars |
+|---|---|---|---|
+| Main API server | `true` | `false` | — |
+| GPU box (NVENC) | `false` | `true` | `MEDIA_VIDEO_ENCODER=av1_nvenc`, `MEDIA_VIDEO_CODEC=av1`, `MEDIA_JIT_VIDEO_ENCODER=h264_nvenc` |
+
+## Web frontend (build-time config)
+
+The web app is a static bundle; its config is baked in **at build time** by Vite (Vite reads the
+repo-root `.env` via `envDir: '../'`). These must be set when running `bun run build`, not at
+runtime:
+
+| Var | Purpose |
+|---|---|
+| `VITE_API_BASE_URL` | API base URL the built frontend calls (e.g. `https://api.ganked.tv`). Falls back to `http://localhost:5050` if unset — so it **must** be set for prod builds. |
+| `VITE_GA_MEASUREMENT_ID` | GA4 measurement id (`G-XXXXXXX`). Analytics is a complete no-op (no script, no cookies) when empty, so it stays off in dev/preview. |
+
+> The Dockerfile build-arg wiring that feeds these into the production web image is part of
+> issue #123. **Cookie-consent gating for analytics is a known follow-up** — GA currently loads
+> whenever the measurement id is present.
+
+## Reference
+
+A full annotated list of every variable lives in [.env.example](.env.example).
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -43,9 +43,11 @@ Migrations are **manual locally** (`make migrate`). In production, set
 `RUN_MIGRATIONS_ON_STARTUP=true` on the migrating instance so a fresh DB self-migrates at boot,
 before serving requests. `/health/ready` only reports healthy once all migrations are applied
 (see [DatabaseMigrator](server/src/GankedTV.Api/Data/DatabaseMigrator.cs) and
-[ReadinessHealthCheck](server/src/GankedTV.Api/Services/Health/ReadinessHealthCheck.cs)). If you run
-more than one API replica, enable the flag on exactly one (or run migrations as a separate
-init step) to avoid concurrent migration races.
+[ReadinessHealthCheck](server/src/GankedTV.Api/Services/Health/ReadinessHealthCheck.cs)). EF Core
+serialises migration runs via a `__EFMigrationsHistory` lock, so multiple replicas booting with the
+flag enabled apply migrations safely (they wait, they don't race). For clarity you may still prefer
+to gate the flag to one replica or run migrations as a separate init step, but it isn't required for
+correctness. The flag accepts `true`/`1`/`yes`/`on`.
 
 ## Health endpoints
 
--- a/server/src/GankedTV.Api/Configuration/ProductionStartupValidator.cs
+++ b/server/src/GankedTV.Api/Configuration/ProductionStartupValidator.cs
@@ -9,6 +9,15 @@ namespace GankedTV.Api.Configuration;
 /// host refuses to boot with a clear, aggregated error rather than running misconfigured and
 /// failing later on the first request. Pure (no I/O) so it is unit-tested directly.
 /// </summary>
+/// <remarks>
+/// The caller in <c>Program.cs</c> passes values read straight from <c>Environment.GetEnvironmentVariable</c>,
+/// deliberately bypassing the DI options binding (which falls back to <c>appsettings</c>/config and
+/// dev defaults like <c>minioadmin</c> or <c>localhost:5173</c>). This enforces an <b>env-only secret
+/// contract</b>: required secrets must arrive via environment variables, per <c>DEPLOYMENT.md</c>.
+/// If a future deployment wires secrets through a config provider instead (Azure Key Vault, AWS
+/// Parameter Store, etc.), this env-only check will report them "missing" — update both the call
+/// site (to read the bound options) and this contract before integrating a vault.
+/// </remarks>
 public static class ProductionStartupValidator
 {
     /// <summary>Dev-default object-storage credentials that must never reach Production.</summary>
--- a/server/src/GankedTV.Api/Configuration/ProductionStartupValidator.cs
+++ b/server/src/GankedTV.Api/Configuration/ProductionStartupValidator.cs
@@ -1,0 +1,76 @@
+using GankedTV.Api.Auth;
+using GankedTV.Api.Services.ObjectStorage;
+
+namespace GankedTV.Api.Configuration;
+
+/// <summary>
+/// Fail-fast validation of required secrets/config when running in Production. Called from
+/// <c>Program.cs</c> after options binding; if <see cref="Validate"/> returns any problems the
+/// host refuses to boot with a clear, aggregated error rather than running misconfigured and
+/// failing later on the first request. Pure (no I/O) so it is unit-tested directly.
+/// </summary>
+public static class ProductionStartupValidator
+{
+    /// <summary>Dev-default object-storage credentials that must never reach Production.</summary>
+    private const string DevDefaultCredential = "minioadmin";
+
+    private const int MinJwtSecretBytes = 32;
+
+    public static IReadOnlyList<string> Validate(
+        string? connectionString,
+        JwtOptions jwt,
+        OAuthOptions oauth,
+        S3Options s3,
+        string? corsOrigins)
+    {
+        var errors = new List<string>();
+
+        if (string.IsNullOrWhiteSpace(connectionString))
+        {
+            errors.Add("DATABASE_URL must be set.");
+        }
+
+        if (string.IsNullOrWhiteSpace(jwt.Secret))
+        {
+            errors.Add("JWT_SECRET must be set.");
+        }
+        else if (System.Text.Encoding.UTF8.GetByteCount(jwt.Secret) < MinJwtSecretBytes)
+        {
+            errors.Add($"JWT_SECRET must be at least {MinJwtSecretBytes} bytes.");
+        }
+
+        if (string.IsNullOrWhiteSpace(oauth.WebOrigin))
+        {
+            errors.Add("WEB_ORIGIN must be set.");
+        }
+
+        if (string.IsNullOrWhiteSpace(corsOrigins))
+        {
+            errors.Add("CORS_ORIGINS must be set.");
+        }
+
+        if (string.IsNullOrWhiteSpace(s3.Endpoint))
+        {
+            errors.Add("S3_ENDPOINT must be set.");
+        }
+
+        if (string.IsNullOrWhiteSpace(s3.AccessKey)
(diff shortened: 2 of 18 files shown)


web/src/router/index.ts
```diff
@@ -144,4 +145,10 @@ router.beforeEach((to) => {
   }
 })
 
+// Emit SPA page views after each successful navigation. No-op until analytics is initialised
+// with a measurement id (production only), so this is inert in dev and tests.
+router.afterEach((to) => {
```
mintycw, line 151: 🔴 **Critical — OAuth tokens leak to Google Analytics**

`to.fullPath` includes the query string. The OAuth callback URL is `/auth/callback?token=<JWT>&refresh=<refresh-token>&returnTo=…`. With `VITE_GA_MEASUREMENT_ID` set, every login will ship the user's JWT *and* refresh token to GA as both `page_path` and `page_location`. Refresh tokens last 30 days, so this is a real third-party credential leak.

This is **independent of the documented cookie-consent follow-up** — consent gating wouldn't help once consent is granted.

**Fix:** strip sensitive query keys (`token`, `refresh`, `code`, `state`, `access_token`, `id_token`) before calling `trackPageView`, or skip tracking on `/auth/callback`. Suggested:

```ts
const SENSITIVE = new Set(['token','refresh','code','state','access_token','id_token'])
router.afterEach((to) => {
  const [path, query = ''] = to.fullPath.split('?')
  const params = new URLSearchParams(query)
  for (const k of [...params.keys()]) if (SENSITIVE.has(k)) params.delete(k)
  const qs = params.toString()
  trackPageView(qs ? `${path}?${qs}` : path)
})
```

Add a regression test that pushes `/auth/callback?token=fake&refresh=fake` and asserts those keys are absent from the `trackPageView` call.

web/src/main.ts
```diff
@@ -28,5 +29,8 @@ configureAuth({
 
 await auth.bootstrap()
 
+// No-op unless VITE_GA_MEASUREMENT_ID is set at build time (production only).
+initAnalytics(import.meta.env.VITE_GA_MEASUREMENT_ID)
```
mintycw, line 33: 🟡 **Important — verify initial page view fires**

`initAnalytics` runs *before* `app.use(router)` on the next line, and `router.afterEach` only handles navigations once the router is installed. Vue Router's behavior on the *initial* route under `createWebHistory` differs from `createMemoryHistory` (which the tests use), so the existing router test at `router/__tests__/index.spec.ts:139-143` doesn't cover this.

If the initial navigation doesn't fire `afterEach`, every session undercounts by 1. Please verify in a real browser, or explicitly call `trackPageView(window.location.pathname + window.location.search)` once `router.isReady()` resolves.

web/src/lib/analytics.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    window.dataLayer = window.dataLayer || []
+    window.gtag = function gtag(...args: unknown[]) {
+      window.dataLayer!.push(args)
+    }
+    window.gtag('js', new Date())
+    // send_page_view:false — SPA page views are emitted by the router (see router/index.ts),
+    // otherwise the initial automatic hit double-counts the first navigation.
+    window.gtag('config', measurementId, { send_page_view: false })
+  }
```
mintycw, line 82: 🟡 **Important (robustness) — `gtag('config', …)` skipped on re-init**

This whole block — including the `gtag('config', measurementId, { send_page_view: false })` call — is guarded by `if (!window.gtag)`. If `initAnalytics` is ever called twice (different measurement id, hot reload, future router re-init), the second call silently drops the config call.

Not a problem today since `initAnalytics` is called exactly once from `main.ts`, but worth either:
- moving `gtag('config', ...)` outside the `if (!window.gtag)` guard, or
- adding a comment noting that re-init is a deliberate no-op for config.

The spec at `analytics.spec.ts:68-73` already anticipates the re-call path, so the assumption is half-codified — just want it explicit.

server/src/GankedTV.Api/Configuration/ProductionStartupValidator.cs
```diff
@@ -0,0 +1,76 @@
...
+        if (string.IsNullOrWhiteSpace(jwt.Secret))
+        {
+            errors.Add("JWT_SECRET must be set.");
+        }
+        else if (System.Text.Encoding.UTF8.GetByteCount(jwt.Secret) < MinJwtSecretBytes)
+        {
+            errors.Add($"JWT_SECRET must be at least {MinJwtSecretBytes} bytes.");
+        }
+
+        if (string.IsNullOrWhiteSpace(oauth.WebOrigin))
```
mintycw, line 51: 🟡 **Important — dev-default screen is incomplete for URL-shaped values**

`minioadmin` is rejected for `S3_ACCESS_KEY` / `S3_SECRET_KEY` (great defense), but the URL-shaped fields are only checked for emptiness:
- `WEB_ORIGIN=http://localhost:5173` would pass.
- `S3_ENDPOINT=localhost:9000` would pass.
- `S3_PUBLIC_URL=http://localhost:9000` would pass.

If you're already screening for `minioadmin`, consider a `RejectLocalhost` helper applied to these three. An operator who copies their `.env` to prod will otherwise boot with dev origins and won't find out until the first cross-origin login fails.

server/src/GankedTV.Api/Configuration/ProductionStartupValidator.cs
```diff
@@ -0,0 +1,76 @@
...
+/// </summary>
+public static class ProductionStartupValidator
+{
+    /// <summary>Dev-default object-storage credentials that must never reach Production.</summary>
+    private const string DevDefaultCredential = "minioadmin";
+
+    private const int MinJwtSecretBytes = 32;
+
+    public static IReadOnlyList<string> Validate(
+        string? connectionString,
```
mintycw, line 29: 🟡 **Important — validator reads raw env, but bindings fall back to `appsettings`/config**

The validator (correctly) bypasses DI-bound dev fallbacks by reading env directly in `Program.cs`. But the actual options binding (`Program.cs:64-65, 79, 252`) does fall back to `builder.Configuration["…"]`. An operator who later wires up a secrets manager (Azure Key Vault, AWS Parameter Store) that binds via config rather than env will get a misleading "missing" error from the validator even though the app would otherwise boot fine.

Today `DEPLOYMENT.md` says "secrets via env only," so this is mostly theoretical — but please add an `<remarks>` to the xmldoc explicitly stating *the validator enforces the env-only contract*, so the next person integrating a vault knows where to look.

server/src/GankedTV.Api/Data/DatabaseMigrator.cs
```diff
@@ -0,0 +1,38 @@
...
+/// self-migrates before the API serves its first request — locally migrations stay manual
+/// (<c>make migrate</c>). Idempotent: a re-run with no pending migrations is a no-op.
+/// </summary>
+public static class DatabaseMigrator
+{
+    /// <summary>Env-var name that gates startup migration. Set to <c>"true"</c> to enable.</summary>
+    public const string EnableEnvVar = "RUN_MIGRATIONS_ON_STARTUP";
+
+    public static bool IsEnabled(string? envValue) =>
+        string.Equals(envValue, "true", StringComparison.OrdinalIgnoreCase);
```
mintycw: 🟢 **Minor — only accepts `"true"`, not `"1"`/`"yes"`**

K8s ConfigMaps and many CI systems use `RUN_MIGRATIONS_ON_STARTUP=1` by convention. A misconfigured deploy will boot silently without migrating; `/health/ready` will stay red indefinitely with a confusing "migrations pending" message.

This is consistent with `bool.TryParse` used elsewhere in `Program.cs:110, 132` (which also rejects `"1"`), so it's not a regression — but worth either accepting `"1"` here, or adding a `<remarks>` in the xmldoc explicitly noting that `"1"` is NOT accepted so the next operator doesn't get bitten.

DEPLOYMENT.md
```diff
@@ -0,0 +1,86 @@
...
+Also recommended in production: `OAUTH_STATE_SECRET` (≥ 32 bytes; required once any OAuth provider
+is configured) and the provider client credentials (`DISCORD_*`, `GOOGLE_*`).
+
+## Startup database migrations
+
+Migrations are **manual locally** (`make migrate`). In production, set
+`RUN_MIGRATIONS_ON_STARTUP=true` on the migrating instance so a fresh DB self-migrates at boot,
+before serving requests. `/health/ready` only reports healthy once all migrations are applied
+(see [DatabaseMigrator](server/src/GankedTV.Api/Data/DatabaseMigrator.cs) and
+[ReadinessHealthCheck](server/src/GankedTV.Api/Services/Health/ReadinessHealthCheck.cs)). If you run
```
mintycw: 🟢 **Minor — concurrent-replica guidance is overcautious**

EF Core's `MigrateAsync` takes a `__EFMigrationsHistory` row lock — concurrent replicas serialise rather than race. So you don't actually need to enable the flag on exactly one replica, although doing so is harmless.

Not wrong, just stricter than necessary. Operators following this literally will add an init container or leader-election step that EF doesn't require. Consider softening to "EF serialises migration runs via a history-table lock, so multiple replicas applying migrations simultaneously is safe — but for clarity you may want to gate it to one replica or run it as a separate init step."

mintycw (commented):
Local review pass — strong server-side hardening overall (validator, migrator, readiness check, dev-token 404, real Testcontainers tests). One Critical finding on the web side that should block merge, plus a few Important/Minor follow-ups inline. Full notes:

**Strengths**
- Health probes correctly ordered before `UseHttpsRedirection`; liveness uses `Predicate = _ => false`.
- `Program.cs` stays thin — all logic in `ProductionStartupValidator` / `DatabaseMigrator` / `ReadinessHealthCheck`, so it stays inside the coverage denominator.
- JWT secret length checked in **bytes** via `Encoding.UTF8.GetByteCount`. `minioadmin` explicitly rejected. Validator reads raw env to bypass dev-fallback bindings.
- `DatabaseMigrator` short-circuits when no pending migrations; both success/failure paths logged.
- Dev-token guard returns **404** (doesn't leak existence).
- Analytics no-op is genuinely no-op; `send_page_view: false` avoids initial-load double-counting; `gtag` is properly typed.
- No `Co-Authored-By: Claude` trailer on the commit.

**Verdict:** Don't merge until Issue #1 (OAuth token leak to GA) is fixed. Everything else can land as follow-up.

server/src/GankedTV.Api/Program.cs
```diff
@@ -425,6 +432,39 @@
...
+        },
+        Environment.GetEnvironmentVariable("CORS_ORIGINS"));
+    if (secretErrors.Count > 0)
+    {
+        throw new InvalidOperationException(
+            "Refusing to start in Production — required configuration is missing or invalid:"
+            + Environment.NewLine
+            + string.Join(Environment.NewLine, secretErrors.Select(e => "  - " + e)));
+    }
+}
```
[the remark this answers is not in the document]
Turbootzz: Skipping — the code already satisfies the cited guideline. The *logic* lives in services: `ProductionStartupValidator.Validate` (pure, 12 unit tests) and `DatabaseMigrator` (tested against a fresh Testcontainer), both inside the coverage denominator. What remains in Program.cs is **config binding** (reading env vars into option objects — which the same guideline explicitly permits) plus a thin fail-fast gate and a scoped `MigrateAsync` call, and that path is exercised end-to-end by `ProductionEnvironmentTests`.

Extracting it into an initializer/IHostedService would (a) add no denominator coverage — the validator is intentionally object-based so it's unit-testable without touching process env, and reading env *inside* a service would make it harder to test; and (b) change fail-fast semantics — the validator currently throws **before `builder.Build()`/`app.Run()`** so the host never starts, whereas an `IHostedService.StartAsync` throw happens later with different wrapping. Running migrations via a scope before `app.Run()` is also the idiomatic ASP.NET pattern. Happy to revisit if the guideline is meant to cover config binding too.

web/src/main.ts
Turbootzz: Verified — no change needed. In Vue Router 4, `afterEach` **does** run for the initial navigation (the first resolved route triggers the global afterEach hooks), so the initial page view fires. The hook is also registered at module-load time in `router/index.ts`, before `app.use(router)` installs the router and kicks off that first navigation, and `initAnalytics` runs before `app.use(router)` in `main.ts` — so the provider is already live when the initial view is emitted.

Because the initial `afterEach` already fires, adding an explicit `trackPageView` on `router.isReady()` would **double-count** the first view, so I'm deliberately not adding it. (`createMemoryHistory` in the tests behaves the same way for afterEach; the existing router test asserts the hook fires on navigation.)

server/src/GankedTV.Api/Configuration/ProductionStartupValidator.cs
Turbootzz: Skipping the localhost screen, for three reasons:

1. **The dominant footgun is already caught.** A copied dev `.env` ships an empty `JWT_SECRET` (rejected: must be ≥32B) and `minioadmin` S3 creds (explicitly rejected), so it can't boot in Production regardless.
2. **localhost is legitimately valid for `S3_ENDPOINT`.** A same-host / sidecar MinIO is reachable at `localhost:9000` from the API container, so a blanket reject would cause false-positive boot failures for a valid topology.
3. **Test-harness coupling.** `AuthApiFactory` (Production path) and `CorsOriginsTests` deliberately use `WEB_ORIGIN=http://localhost:5173` — `Production_DoesNotAutoAllowLocalhost` specifically asserts that localhost origin is allowed *because it's the WebOrigin*. Rejecting localhost on `WEB_ORIGIN`/`S3_PUBLIC_URL` would break those tests' premise and require reworking a well-designed harness.

Net: the marginal extra coverage doesn't justify the false-positive risk + harness churn here. If we want this later, the cleaner place is a startup *warning* (not a hard fail) for localhost in browser-facing URLs.

mintycw (approved):
Approving — no blockers. Reviewed the full diff (base df73fd8 → head cd08fcf).

**Verified end-to-end:**
- Health endpoints are mapped as terminal middleware **before** `UseHttpsRedirection` (Program.cs:517-526) — HTTP probes get 200/503, not 307. ✅
- OAuth-token redaction is genuinely complete, not papered over: `/auth/callback` is query-string (not fragment), `trackPageView` has a single sink via `router.afterEach`, and the redacted-key set covers both auth-code (`code`/`state`) and implicit (`access_token`/`id_token`) on top of in-product `token`/`refresh`. ✅
- `JWT_SECRET` length check uses `Encoding.UTF8.GetByteCount` (real bytes, not chars). ✅
- `/dev/token` runtime guard returns 404, not 403 — indistinguishable from "not present". ✅
- `Program.cs` stays pure DI wiring; real logic in `ProductionStartupValidator` / `ReadinessHealthCheck` / `DatabaseMigrator` so the coverage-exclusion contract holds. ✅
- `DatabaseMigrator.IsEnabled` accepts `1`/`yes`/`on` — silently-skipped migrations are a documented foot-gun, this is the right fix. ✅
- `analytics.ts` re-init is correct: script bootstrap guarded by `!window.gtag`, but `gtag('config', …)` runs every call. ✅

**Non-blocking follow-ups worth filing as issues** (none of these need to land here):

1. **`DatabaseMigrator` disabled-path is silent.** No log line when `RUN_MIGRATIONS_ON_STARTUP` is unset/false on a fresh prod DB — the only signal is `/health/ready` returning 503. Add an else-branch log so ops aren't debugging a stuck-red probe blind.
2. **`ReadinessHealthCheck` re-runs `GetPendingMigrationsAsync` every probe** (ReadinessHealthCheck.cs:25). Reads `__EFMigrationsHistory` + migration-assembly metadata on every k8s probe × N pods. Memoize after first success; keep `CanConnectAsync` per-call.
3. **`appsettings.Production.json` defaults assume multi-host topology** (`TranscodeWorkerEnabled: false`). Single-box deploys will queue media jobs forever with no worker. Either flip the default or emit a startup warning when the toggle combo looks misconfigured.
4. **Validator integration test only exercises the happy path.** `AuthApiFactory` pre-sets every required env, so a regression that drops the `IsProduction()` guard or moves the call after `app.Run()` wouldn't fail any test. Add one `WebApplicationFactory`-based test asserting boot fails with a missing `JWT_SECRET`.
5. **Validator throws a multi-line `InvalidOperationException`.** Some log shippers truncate at the first newline. Consider emitting each error as its own log entry before throwing.

**Minor polish (skip-or-keep, your call):**
- `analytics.ts` `stripSensitiveParams` drops URL hash fragments when rebuilding the path — not a leak, just makes GA page paths differ from the address bar.
- `DEPLOYMENT.md` could clarify "32 bytes from any source" so base64 vs hex generation isn't ambiguous.
- Validator constructor takes `JwtOptions`/`OAuthOptions`/`S3Options` only to populate 3 fields each, with throwaway `"n/a"` fillers — a flat env-string record would read cleaner.

Scope matches #122 exactly. Nothing from #123 (containerization) snuck in. Ship it.

